Issue #5, opened 2023-01-02
Regression on 3D visualization


mofeing:
Plotting directly in 3D using `Spring(dim=3)` from `NetworkLayout` no longer works (it plots in 3D).

```julia
using Tenet
using Makie
using GLMakie
using NetworkLayout

tn = rand(GenericTensorNetwork, 60, 3)

plot(tn, layout=Spring(dim=3))
```

PR #19, opened 2023-02-09
Refactor visualization functions

jofrevalles:
Plot recipes currently describe plot primitives that live in an `Axis` but can not change the `Axis` parameters. Therefore, with plot recipes, we can't currently include `Colorbar`s, add interactions, and in general, customize the `Axis`.
In this PL we address #5, and we change the plot function so now it can take a `GridPosition` and plot into it. This enabled us to add the following features:

- Added integration to plot the contraction paths into subplots.
- Added the possibility to create 3D plots.

### Example:
```julia
julia> using Tenet
julia> using Makie
julia> using GLMakie
julia> tn = rand(TensorNetwork, 60, 3)
TensorNetwork{Arbitrary}(#tensors=60, #inds=90)
julia> f = Figure(backgroundcolor="ivory", resolution=(1000, 700))
julia> scatter(f[1,1], 1..4, log)
Makie.AxisPlot(Axis (1 plots), Scatter{Tuple{Vector{Point{2, Float32}}}})
julia> lines(f[2,1], 1..10, cos)
Makie.AxisPlot(Axis (1 plots), Lines{Tuple{Vector{Point{2, Float32}}}})
julia> using NetworkLayout
julia> plot!(f[1:2,2], tn; layout=Spring(dim=3))
(Combined{GraphMakie.graphplot, Tuple{Graphs.SimpleGraphs.SimpleGraph{Int64}}}, LScene())
```
image

### DIFF
--- a/src/Visualization.jl
+++ b/src/Visualization.jl
@@ -1,23 +1,46 @@
 using Graphs: SimpleGraph, Edge, nv
-using GraphMakie: graphplot!, GraphPlot
+using GraphMakie: graphplot!, GraphPlot, to_colormap, get_node_plot
 using Combinatorics: combinations
+using GraphMakie.NetworkLayout: IterativeLayout
 import Makie
 
-Makie.plottype(::TensorNetwork) = GraphPlot
+const MAX_NODE_SIZE = 20.
 
-function Makie.plot!(P::GraphPlot{Tuple{TensorNetwork{A}}}; kwargs...) where {A<:Ansatz}
-    tn = P[1][]
+function Makie.plot(tn::TensorNetwork{A}; kwargs...) where {A<:Ansatz}
+    f = Makie.Figure()
+
+    p, ax = Makie.plot!(f[1,1], tn; kwargs...)
+    display(f)
+
+    return f, ax, p
+end
+
+function Makie.plot!(f::Makie.GridPosition, tn::TensorNetwork{A}; kwargs...) where {A<:Ansatz}
+    scene = Makie.Scene()
+    default_attrs = Makie.default_theme(scene, GraphPlot)
+
+    kwargs = Dict{Symbol,Any}(kwargs)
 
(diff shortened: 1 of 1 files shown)


src/Visualization.jl
```diff
@@ -1,23 +1,46 @@
...
-        kwargs[:node_size] = [max(10, log2(size(tensors(tn, i)) |> prod)) for i in 1:nv(graph)]
+
+    pos = IdDict(tensor => i for (i, tensor) in enumerate(tensors(tn)))
+    graph = SimpleGraph([Edge(pos[a], pos[b]) for ind in inds(tn) for (a, b) in combinations(links(ind), 2)])
+
+    log_size = [log2(size(tensors(tn, i)) |> prod) for i in 1:nv(graph)]
+
+    min_size, max_size = extrema(log_size)
+
+    kwargs[:node_size] = (log_size/max_size) * MAX_NODE_SIZE
```
mofeing, line 31: I was quit satisfied with the previous formula. Node size should not be normalized to min-max sizes of tensors. If a node is small, it's small because the tensor is small and the intuition should work that way.

```suggestion
    kwargs[:node_size] = [max(10, log2(size(tensors(tn,i)) |> prod)) for i in 1:nv(graph)]
```

src/Visualization.jl
```diff
@@ -1,23 +1,46 @@
 using Graphs: SimpleGraph, Edge, nv
-using GraphMakie: graphplot!, GraphPlot
+using GraphMakie: graphplot!, GraphPlot, to_colormap, get_node_plot
 using Combinatorics: combinations
+using GraphMakie.NetworkLayout: IterativeLayout
 import Makie
 
-Makie.plottype(::TensorNetwork) = GraphPlot
+const MAX_NODE_SIZE = 20.
```
mofeing, line 8:
```suggestion
```

mofeing (changes requested):

PR #28, opened 2023-02-22
Fix from last PR: bug in plot functions

jofrevalles:
Small fix from last PR, the plot function crashed since the function `labels` conflicts with the keyword argument `labels`.


mofeing:
This bug wasn't catch by tests because we don't test visualization. Maybe we should think about testing that plotting doesn't crash?

jofrevalles:
Yep, my fault. Since this was added by code suggestion, I did not test that. I will next time

mofeing:
Maybe yeah. The only thing is that CI runs do not have a graphical interface so they probably will fail. I think there is a package for creating a fake graphical interface just for this kind of things.

PR #30, opened 2023-02-22
Fix crash in `transform` function: `transformations` argument

jofrevalles:
### Summary
This PR addresses the bug found in the `transform` function. This bug was not found before because this function is not included in the tests at the moment.
Previously, we could not call the `transform` function since the `transformations` argument accepted the type `Sequence{Transformation}` and `Transformation` is an abstract type. This is now changed so it accepts a `Sequence` of `struct`s that inherit from `Transformation`.

### Example
```julia
julia> using Tenet
julia> tn = TensorNetwork([Tensor(rand([2, 2, 2]...), tuple([:i, :j, :k]...)), Tensor(rand([2, 2, 2]...), tuple([:i, :k, :l]...)), Tensor(rand([2, 2, 2]...), tuple([:i, :l, :j]...))])
julia> tn = transform(tn, [Tenet.HyperindConverter()])
TensorNetwork{Arbitrary}(#tensors=4, #inds=6)
```
Although this may be counter-intuitive since the `transform!` function is called with:
```julia
julia> transform!(tn, Tenet.HyperindConverter)
```
What do you think, @mofeing ?

### DIFF
--- a/src/Transformations.jl
+++ b/src/Transformations.jl
@@ -4,14 +4,14 @@ abstract type Transformation end
 
 function transform! end
 
-function transform!(tn::TensorNetwork, transformations::Sequence{Transformation})
+function transform!(tn::TensorNetwork, transformations::Sequence{A}) where {A<:Transformation}
     for transformation in transformations
-        transform!(tn, transformation)
+        transform!(tn, typeof(transformation))
     end
     return tn
 end
 
-function transform(tn::TensorNetwork, transformations::Sequence{Transformation})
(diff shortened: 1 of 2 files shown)


mofeing:
The way you have written the code won't work because it implies that the type of all the elements in a `Sequence` (which is just a helper alias for `Union{Tuple,Vector}`.

Also, the PR fails an Aqua test.

I've refactored the code in such a way that it solves these problems and also abstract the call to `transform!` in a better way.
Also, I've added some tests for theses cases.

Issue #32, opened 2023-02-24
Open indices are not shown on plots


mofeing:
Open indices are not shown on plots. This is due to the generation of the graph in the `plot!` method, because `Graphs.SimpleGraph` does not support open edges.

One solution is to introduces mock nodes with size 0 so edges can be created and represented in the plots.

PR #34, opened 2023-03-03
Add support for visualization of open indices in plot functions

jofrevalles:
### Summary
In this PR we address issue #32: we fix the plot functions so now we can visualize the open indices that a `TensorNetwork` may have. This is done by adding 'ghost' tensors of size zero at the edge of each open index.

### Example
```julia
julia> using Tenet
julia> using GLMakie; using Makie
julia> tn = TensorNetwork([Tensor(rand([2, 2, 2, 2, 2]...), tuple([:i, :t, :k, :x, :y]...)), Tensor(rand([2, 2, 2, 2]...), tuple([:i, :k, :s, :p]...)), Tensor(rand([2, 2]...), tuple([:i, :s]...)), Tensor(rand([2, 2, 2]...), tuple([:p, :t, :z]...))])
TensorNetwork{Arbitrary}(#tensors=4, #inds=8)
julia> using NetworkLayout
julia> plot(tn; labels=true, layout=Spring(dim=3))
Tuple{Symbol, Vararg{Symbol}}[(:p, :t, :z), (:i1, :t, :k, :x, :y), (:i2, :k, :s, :p), (:i3, :s), (:i1, :i2, :i3), (:y, :x), (:z,)]
julia> [tensor.labels for tensor in tn.tensors]
4-element Vector{Tuple{Symbol, Symbol, Vararg{Symbol}}}:
 (:i, :t, :k, :x, :y)
 (:i, :k, :s, :p)
 (:i, :s)
 (:p, :t, :z)
julia> openinds(tn)
3-element Vector{Index}:
 :y
 :z
 :x
```
image

### DIFF
--- a/src/Visualization.jl
+++ b/src/Visualization.jl
@@ -1,5 +1,5 @@
-using Graphs: SimpleGraph, Edge, edges, ne, nv
-using GraphMakie: graphplot!, GraphPlot, to_colormap, get_node_plot, add_edge!, add_vertex!
+using Graphs: SimpleGraph, Edge, edges, ne, nv, add_edge!, add_vertex!
+using GraphMakie: graphplot!, GraphPlot, to_colormap, get_node_plot
 using Combinatorics: combinations
 using GraphMakie.NetworkLayout: IterativeLayout
 import Makie
--- a/src/Visualization.jl
+++ b/src/Visualization.jl
@@ -1,4 +1,4 @@
-using Graphs: SimpleGraph, Edge, edges, ne, nv, add_edge!, add_vertex!
+using Graphs: SimpleGraph, Edge, edges, ne, nv, add_edge!, add_vertex!, src, dst
 using GraphMakie: graphplot!, GraphPlot, to_colormap, get_node_plot
 using Combinatorics: combinations
 using GraphMakie.NetworkLayout: IterativeLayout
@@ -28,9 +28,9 @@ function Makie.plot!(f::Makie.GridPosition, tn::TensorNetwork{A}; labels = false
     # TODO recognise them by using `DeltaArray` or `Diagonal` representations
     copytensors = findall(t -> haskey(t.meta, :dual), tensors(tn))
 
-    opentensors = findall(t -> any(s -> s ∈ (o -> nameof(o)).(openinds(tn)), t.labels), tensors(tn))
+    opentensors = findall(t-> !isempty(labels(t) ∩ openinds(tn)), tensors(tn))
 
-    opencounter = Dict(tensor => 1 for tensor in opentensors)
+    opencounter = IdDict(tensor => 1 for tensor in opentensors)
     ghostnodes = map(openinds(tn)) do ind
         add_vertex!(graph)
         node = nv(graph) # TODO is this the best way to get the id of the newly created node?
@@ -51,17 +51,18 @@ function Makie.plot!(f::Makie.GridPosition, tn::TensorNetwork{A}; labels = false
             copies = filter((x -> x ∈ copytensors), [edge.src, edge.dst])
             notghosts = filter((x -> x ∉ ghostnodes), [edge.src, edge.dst])
 
+            # TODO refactor this code
             if length(notghosts) == 2 # there are no ghost nodes in this edge
                 if isempty(copies) # there are no copy tensors in the nodes of this edge
-                    push!(elabels, join(Tenet.labels(tensors(tn)[edge.src]) ∩ Tenet.labels(tensors(tn)[edge.dst]), ','))
+                    push!(elabels, join(Tenet.labels(tensors(tn)[src(edge)]) ∩ Tenet.labels(tensors(tn)[dst(edge)]), ','))
                     push!(elabels_color, :black)
                 else
                     push!(elabels, string(tensors(tn)[copies[]].meta[:dual]))
                     push!(elabels_color, :grey)
                 end
             else
-                tensor_oinds = filter(id -> nameof(id) ∈ tensors(tn)[only(notghosts)].labels, openinds(tn))
-                indices = (id -> nameof(id)).(tensor_oinds)
+                tensor_oinds = filter(id -> nameof(id) ∈ labels(tensors(tn)[only(notghosts)]), openinds(tn))
+                indices = nameof.(tensor_oinds)
 
                 push!(elabels, string(indices[opencounter[only(notghosts)]]))
                 push!(elabels_color, :black)
--- a/src/Visualization.jl
+++ b/src/Visualization.jl
@@ -28,7 +28,7 @@ function Makie.plot!(f::Makie.GridPosition, tn::TensorNetwork{A}; labels = false
     # TODO recognise them by using `DeltaArray` or `Diagonal` representations
     copytensors = findall(t -> haskey(t.meta, :dual), tensors(tn))
 
-    opentensors = findall(t-> !isempty(Tenet.labels(t) ∩ (i -> nameof(i)).(openinds(tn))), tensors(tn))
+    opentensors = findall(t-> !isempty(Tenet.labels(t) ∩ nameof.(openinds(tn))), tensors(tn))
 
     opencounter = IdDict(tensor => 1 for tensor in opentensors)
     ghostnodes = map(openinds(tn)) do ind
--- a/src/Visualization.jl
+++ b/src/Visualization.jl
@@ -28,19 +28,14 @@ function Makie.plot!(f::Makie.GridPosition, tn::TensorNetwork{A}; labels = false
     # TODO recognise them by using `DeltaArray` or `Diagonal` representations
     copytensors = findall(t -> haskey(t.meta, :dual), tensors(tn))
 
-    oinds = openinds(tn)
-    ghostnodes = []
-    while !isempty(oinds)
-        for ind in oinds
-            tensor = links(ind)[1]
-            tensor_oinds = filter(id -> id.name ∈ tensor.labels, oinds)
+    ghostnodes = map(openinds(tn)) do ind
+        add_vertex!(graph)
 
-            add_vertex!(graph)
-            add_edge!(graph, pos[tensor], nv(graph))
+        node = nv(graph) # TODO is this the best way to get the id of the newly created node?
+        tensor = only(links(ind))
+        add_edge!(graph, node, pos[tensor])
 
-            push!(ghostnodes, nv(graph))
-            filter!(id -> id ∉ tensor_oinds, oinds)
-        end
+        return node
     end
 
     kwargs[:node_size] = [i ∈ ghostnodes ? 0 : max(15, log2(size(tensors(tn, i)) |> prod)) for i in 1:nv(graph)]
@@ -72,7 +67,6 @@ function Makie.plot!(f::Makie.GridPosition, tn::TensorNetwork{A}; labels = false
             end
         end
     end
-
 
     if haskey(kwargs, :layout) && kwargs[:layout] isa IterativeLayout{3}
         ax = Makie.LScene(f[1, 1])
--- a/src/Visualization.jl
+++ b/src/Visualization.jl
@@ -28,7 +28,7 @@ function Makie.plot!(f::Makie.GridPosition, tn::TensorNetwork{A}; labels = false
     # TODO recognise them by using `DeltaArray` or `Diagonal` representations
     copytensors = findall(t -> haskey(t.meta, :dual), tensors(tn))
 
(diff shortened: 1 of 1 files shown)


src/Visualization.jl
```diff
@@ -22,13 +22,28 @@ function Makie.plot!(f::Makie.GridPosition, tn::TensorNetwork{A}; labels = false
...
+            filter!(id -> id ∉ tensor_oinds, oinds)
+        end
+    end
+
     pos = IdDict(tensor => i for (i, tensor) in enumerate(tensors(tn)))
     graph = SimpleGraph([Edge(pos[a], pos[b]) for ind in inds(tn) for (a, b) in combinations(links(ind), 2)])
 
     # TODO recognise them by using `DeltaArray` or `Diagonal` representations
     copytensors = findall(t -> haskey(t.meta, :dual), tensors(tn))
+    outtensors = findall(t -> haskey(t.meta, :out), tensors(tn))
```
mofeing, line 44: Can you do add the output _ghost_ nodes in the `graph` and not create random tensors?

src/Visualization.jl
```diff
@@ -1,5 +1,5 @@
 using Graphs: SimpleGraph, Edge, edges, ne, nv
-using GraphMakie: graphplot!, GraphPlot, to_colormap, get_node_plot
+using GraphMakie: graphplot!, GraphPlot, to_colormap, get_node_plot, add_edge!, add_vertex!
```
mofeing, line 2: `add_edge!` and `add_vertex!` are part of the `Graphs` package, not `GraphMakie`

src/Visualization.jl
```diff
@@ -28,27 +28,52 @@ function Makie.plot!(f::Makie.GridPosition, tn::TensorNetwork{A}; labels = false
...
+                tensor_oinds = filter(id -> id.name ∈ tensors(tn)[notghosts[1]].labels, openinds(tn))
+                indices = (id -> id.name).(tensor_oinds)
+
+                push!(elabels, join(indices, ","))
+                push!(elabels_color, :black)
             end
         end
     end
 
+
```
mofeing, line 76: This space is redundant
```suggestion
```

src/Visualization.jl
```diff
@@ -28,27 +28,52 @@ function Makie.plot!(f::Makie.GridPosition, tn::TensorNetwork{A}; labels = false
...
     copytensors = findall(t -> haskey(t.meta, :dual), tensors(tn))
 
-    kwargs[:node_size] = [max(15, log2(size(tensors(tn, i)) |> prod)) for i in 1:nv(graph)]
-    kwargs[:node_marker] = [i ∈ copytensors ? :diamond : :circle for i in 1:length(tensors(tn))]
-    kwargs[:node_color] = [i ∈ copytensors ? :black : :white for i in 1:length(tensors(tn))]
+    oinds = openinds(tn)
+    ghostnodes = []
+    while !isempty(oinds)
+        for ind in oinds
+            tensor = links(ind)[1]
```
mofeing, line 35: Better use `only`
```suggestion
            tensor = only(links(ind))
```

src/Visualization.jl
```diff
@@ -28,27 +28,52 @@ function Makie.plot!(f::Makie.GridPosition, tn::TensorNetwork{A}; labels = false
...
+            tensor_oinds = filter(id -> id.name ∈ tensor.labels, oinds)
```
mofeing, line 36: Use the getter methods to access the fields
```suggestion
            tensor_oinds = filter(id -> nameof(id) ∈ labels(tensor), oinds)
```

src/Visualization.jl
```diff
@@ -28,27 +28,52 @@ function Makie.plot!(f::Makie.GridPosition, tn::TensorNetwork{A}; labels = false
...
+
+            add_vertex!(graph)
+            add_edge!(graph, pos[tensor], nv(graph))
+
+            push!(ghostnodes, nv(graph))
+            filter!(id -> id ∉ tensor_oinds, oinds)
+        end
+    end
```
mofeing, line 44: In general, I don't see why the `while` loop is needed, or why the need for filtering. I think the following code is cleaner and much more straightforward.
```suggestion
    ghostnodes = map(openinds(tn)) do ind
        add_vertex!(graph)

        node = nv(graph) # TODO is this the best way to get the id of the newly created node?
        tensor = only(links(ind))
        add_edge!(graph, node, pos[tensor])

        return node
    end
```

src/Visualization.jl
```diff
@@ -28,27 +28,52 @@ function Makie.plot!(f::Makie.GridPosition, tn::TensorNetwork{A}; labels = false
...
+                    push!(elabels_color, :grey)
+                end
             else
-                push!(elabels, string(tensors(tn)[copies[]].meta[:dual]))
-                push!(elabels_color, :grey)
+                tensor_oinds = filter(id -> id.name ∈ tensors(tn)[notghosts[1]].labels, openinds(tn))
+                indices = (id -> id.name).(tensor_oinds)
+
+                push!(elabels, join(indices, ","))
+                push!(elabels_color, :black)
```
mofeing, line 71: Could you explain me this code? I guess it adds the `labels` to the plot but Im not sure of some steps.

mofeing (changes requested):
I don't understand some steps. Could you modify some code or explain it to me a lil?

src/Visualization.jl
jofrevalles: This part of the code it checks the two nodes at each vertex. If there are no `ghostnode`s in the nodes, it `join`s the labels. If there is one `ghostnode` (the only other possibility is this one) it takes the label of the `notghost`.

src/Visualization.jl
jofrevalles: Okey, this can also work. The `while` and `for` was there since that worked better when I created random tensors. I agree that this is better now.

src/Visualization.jl
jofrevalles: True, fixed.

jofrevalles (commented):
I agree with your comments. I committed them now

src/Visualization.jl
mofeing: I don't think this is a good idea. This index union assumes both indices will be connected to the same tensor, but its something that we don't know yet.

I think leaving them separate is much better.

src/Visualization.jl
mofeing: I'm gonna restate the same as above. I don't think joining open indices is a good idea. Actually, from watching the plot you should be able to count the amount of open indices but if you join indices, then you cannot or it's much more difficult. Open indices should remain separate.

src/Visualization.jl
jofrevalles: Okay, I joined the labels because we do this when we have tensors that share indices. I will change the code to work as you say.

jofrevalles:
Okay, now it works as expected:
```julia
julia> using Tenet
julia> using GLMakie; using Makie
julia> tn = TensorNetwork([Tensor(rand([2, 2, 2, 2]...), tuple([:x, :y, :z, :t]...)), Tensor(rand([2, 2]...), tuple([:x, :y]...)), Tensor(rand([2]...), tuple([:x]...))])
julia> plot(tn; labels=true)
```
image

I have added a dictionary named `opencounter` which counts the iteration of each open index so we can add each label consecutively.

src/Visualization.jl
```diff
@@ -28,23 +28,44 @@ function Makie.plot!(f::Makie.GridPosition, tn::TensorNetwork{A}; labels = false
...
+                    push!(elabels_color, :black)
+                else
+                    push!(elabels, string(tensors(tn)[copies[]].meta[:dual]))
+                    push!(elabels_color, :grey)
+                end
             else
-                push!(elabels, string(tensors(tn)[copies[]].meta[:dual]))
-                push!(elabels_color, :grey)
+                tensor_oinds = filter(id -> nameof(id) ∈ tensors(tn)[only(notghosts)].labels, openinds(tn))
+                indices = (id -> nameof(id)).(tensor_oinds)
```
mofeing, line 64:
```suggestion
                indices = nameof.(tensor_oinds)
```

src/Visualization.jl
```diff
@@ -28,23 +28,44 @@ function Makie.plot!(f::Makie.GridPosition, tn::TensorNetwork{A}; labels = false
...
         elabels_color = Vector{Symbol}([])
 
         for edge in edges(graph)
             copies = filter((x -> x ∈ copytensors), [edge.src, edge.dst])
+            notghosts = filter((x -> x ∉ ghostnodes), [edge.src, edge.dst])
 
-            if isempty(copies) # there are no copy tensors in the nodes of this edge
-                push!(elabels, join(Tenet.labels(tensors(tn)[edge.src]) ∩ Tenet.labels(tensors(tn)[edge.dst]), ','))
-                push!(elabels_color, :black)
+            if length(notghosts) == 2 # there are no ghost nodes in this edge
```
mofeing, line 54:
```suggestion
            # TODO refactor this code
            if length(notghosts) == 2 # there are no ghost nodes in this edge
```

src/Visualization.jl
```diff
@@ -28,23 +28,44 @@ function Makie.plot!(f::Makie.GridPosition, tn::TensorNetwork{A}; labels = false
...
+                if isempty(copies) # there are no copy tensors in the nodes of this edge
+                    push!(elabels, join(Tenet.labels(tensors(tn)[edge.src]) ∩ Tenet.labels(tensors(tn)[edge.dst]), ','))
```
mofeing, line 56:
```suggestion
                    push!(elabels, join(Tenet.labels(tensors(tn)[src(edge)]) ∩ Tenet.labels(tensors(tn)[dst(edge)]), ','))
```

src/Visualization.jl
```diff
@@ -28,23 +28,44 @@ function Makie.plot!(f::Makie.GridPosition, tn::TensorNetwork{A}; labels = false
...
+                    push!(elabels_color, :black)
+                else
+                    push!(elabels, string(tensors(tn)[copies[]].meta[:dual]))
+                    push!(elabels_color, :grey)
+                end
             else
-                push!(elabels, string(tensors(tn)[copies[]].meta[:dual]))
-                push!(elabels_color, :grey)
+                tensor_oinds = filter(id -> nameof(id) ∈ tensors(tn)[only(notghosts)].labels, openinds(tn))
```
mofeing, line 63:
```suggestion
                tensor_oinds = filter(id -> nameof(id) ∈ labels(tensors(tn)[only(notghosts)]), openinds(tn))
```

src/Visualization.jl
```diff
@@ -1,4 +1,4 @@
-using Graphs: SimpleGraph, Edge, edges, ne, nv
+using Graphs: SimpleGraph, Edge, edges, ne, nv, add_edge!, add_vertex!
```
mofeing, line 1:
```suggestion
using Graphs: SimpleGraph, Edge, edges, ne, nv, add_edge!, add_vertex!, src, dst
```

mofeing (changes requested):
Some minor corrections and I'll merge it. In general is great, but I'm not convinced about the `opencounter` variable and the `if...else...` structure in the end.
Any way, let's refactor it in the future.

mofeing (approved):

src/Visualization.jl
```diff
@@ -28,23 +28,45 @@ function Makie.plot!(f::Makie.GridPosition, tn::TensorNetwork{A}; labels = false
...
+    kwargs[:node_size] = [i ∈ ghostnodes ? 0 : max(15, log2(size(tensors(tn, i)) |> prod)) for i in 1:nv(graph)]
+    kwargs[:node_marker] = [i ∈ copytensors ? :diamond : :circle for i in 1:nv(graph)]
+    kwargs[:node_color] = [i ∈ copytensors ? :black : :white for i in 1:nv(graph)]
 
     if labels
         elabels = Vector{String}([])
         elabels_color = Vector{Symbol}([])
 
         for edge in edges(graph)
             copies = filter((x -> x ∈ copytensors), [edge.src, edge.dst])
```
mofeing, line 51:
```suggestion
            copies = filter(x -> x ∈ copytensors, [edge.src, edge.dst])
```

src/Visualization.jl
```diff
@@ -28,23 +28,45 @@ function Makie.plot!(f::Makie.GridPosition, tn::TensorNetwork{A}; labels = false
...
+            notghosts = filter((x -> x ∉ ghostnodes), [edge.src, edge.dst])
```
mofeing, line 52:
```suggestion
            notghosts = filter(x -> x ∉ ghostnodes, [edge.src, edge.dst])
```

mofeing (approved):

PR #36, opened 2023-03-14
Fix label permutation on `MatrixProductState`

jofrevalles:
### Summary
In `MatrixProductState`, we first have the labels sorted in `(:l, :r, :i, :o)` and then we want to have it in `order`, but the permutator permutes from `order` to `(:l, :r, :i, :o)`, so instead of `permute!` we need `invpermute!`. This didn't fail in tests because in the case we considered the permutation was the same as the inverse permutation.

### Example
This code failed without this PR, and now works properly.
```julia
julia> using Tenet; using Test
julia> @test begin
           arrays = [rand(3, 1), rand(3, 1, 3), rand(1, 3)]
           MatrixProductState{Open}(arrays, order = (:r, :p, :l)) isa TensorNetwork{MatrixProductState{Open}}
       end
Test Passed
```

### DIFF
--- a/src/MatrixProductState.jl
+++ b/src/MatrixProductState.jl
@@ -49,7 +49,7 @@ function MatrixProductState{Open}(arrays; χ = nothing, order = (:l, :r, :p), me
         rind = vinds[(i, mod1(i + 1, n))]
 
         labels = [lind, rind, pinds[i]]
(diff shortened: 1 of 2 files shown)


mofeing:
Thanks! Can you add some tests for this?

Issue #29, opened 2023-02-22
Test plotting on CI


mofeing:
From #28, we found out that there might be crashes on plotting code. We should not seek testing deterministic behaviour but just checking that it does not crash.

Problem of writing a test that calls `Makie.plot` is that it will fail on CI runs because they do not have a graphical user interface. I recall that there was a package for creating a fake graphic server, to be used in these cases. I don't remember the name but probably is used by GraphMakie or some Makie backend.

PR #37, opened 2023-03-16
Add support for tests in visualization functions

jofrevalles:
From #29

### Summary
With the previous CI, we could not add tests to ensure that the plotting functions (`plot`, `plot!`) worked without errors. This was because the CI used did not have a graphical interface. In this PR we fix this and we also add tests for `Visualization.jl`.


.github/workflows/CI.yml
```diff
@@ -41,15 +41,19 @@ jobs:
           version: ${{ matrix.version }}
           arch: ${{ matrix.arch }}
       - uses: julia-actions/cache@v1
-        with:
-          cache-name: CI - Julia ${{ matrix.version }} - ${{ matrix.os }} - ${{ matrix.arch }}
-          cache-compiled: true
+      - run: sudo apt-get update && sudo apt-get install -y xorg-dev mesa-utils xvfb libgl1 freeglut3-dev libxrandr-dev libxinerama-dev libxcursor-dev libxi-dev libxext-dev
```
mofeing, line 44: Why did you delete this?

jofrevalles: I think I wanted to simply add the new one, I can add it again, sorry.

test/Visualization_test.jl
```diff
@@ -0,0 +1,32 @@
+@testset "Visualization" begin
+    using CairoMakie
+    using Tenet
+    using NetworkLayout: Spring
+
+    @testset "`plot``" begin
```
mofeing, line 6:
```suggestion
    @testset "plot`" begin
```

test/Visualization_test.jl
```diff
@@ -0,0 +1,32 @@
...
+        tn = TensorNetwork([
+            Tensor(rand([2, 2, 2, 2]...), tuple([:x, :y, :z, :t]...)),
+            Tensor(rand([2, 2]...), tuple([:x, :y]...)),
+            Tensor(rand([2]...), tuple([:x]...))
+            ])
```
mofeing, line 11:
```suggestion
        tn = TensorNetwork([
            Tensor(rand(2, 2, 2, 2), (:x, :y, :z, :t)),
            Tensor(rand(2, 2), (:x, :y)),
            Tensor(rand(2), (:x))
            ])
```

src/Visualization.jl
```diff
@@ -8,15 +8,14 @@ function Makie.plot(tn::TensorNetwork{A}; kwargs...) where {A}
...
-    # return f, ax, p
-    return f
+    return f, ax, p
 end
 
 function Makie.plot!(f::Makie.GridPosition, tn::TensorNetwork{A}; labels = false, kwargs...) where {A}
-    # scene = Makie.Scene()
-    # default_attrs = Makie.default_theme(scene, GraphPlot)
+    scene = Makie.Scene()
+    default_attrs = Makie.default_theme(scene, GraphPlot)
```
mofeing, line 18: These variables are not used.
```suggestion
    # scene = Makie.Scene()
    # default_attrs = Makie.default_theme(scene, GraphPlot)
```

Actually, we could delete them.

src/Visualization.jl
```diff
@@ -8,15 +8,14 @@ function Makie.plot(tn::TensorNetwork{A}; kwargs...) where {A}
     f = Makie.Figure()
 
     p, ax = Makie.plot!(f[1, 1], tn; kwargs...)
-    # display(f)
+    display(f)
 
-    # return f, ax, p
-    return f
+    return f, ax, p
```
mofeing, line 13: This code gave me problems like this. That's way I left it the other way.

mofeing: Are you sure of the call to `display`? Other `plot` methods do not return the `ax` object.

.github/workflows/CI.yml
mofeing: Do we really need all those dependencies?

.github/workflows/CI.yml
```diff
@@ -41,15 +41,19 @@ jobs:
...
           using Pkg
           pkg"registry add https://github.com/bsc-quantic/Registry.git"
           pkg"registry add General"
         shell: julia {0}
+      - name: Install Julia dependencies
+        shell: julia --project=monorepo {0}
+        run: |
+          using Pkg;
+          # dev mono repo versions
+          pkg"dev . "
```
mofeing, line 56: Why this? Actually to install Julia dependencies you would call `Pkg.instantiate` which is automatically called when running the tests.
```suggestion
```

mofeing (changes requested):

.github/workflows/CI.yml
```diff
@@ -41,15 +41,19 @@ jobs:
           version: ${{ matrix.version }}
           arch: ${{ matrix.arch }}
       - uses: julia-actions/cache@v1
```
mofeing, line 43:
```suggestion
      - uses: julia-actions/cache@v1
        with:
           cache-name: CI - Julia ${{ matrix.version }} - ${{ matrix.os }} - ${{ matrix.arch }}
           cache-compiled: true
```

mofeing:
> I also want to add this to `Quac.jl` and `OptimizedEinsum.jl`.

For `OptimizedEinsum` makes sense. For `Quac` no because it's simply SVG code.

mofeing:
Modify the `codecov.yml` file so code coverage stops ignoring the `src/Visualization.jl` file.

.github/workflows/CI.yml
jofrevalles: I don't think so, sorry. This is for the GLMakie backend. I will remove this.

src/Visualization.jl
jofrevalles: I always have found that `plot` returns `f, ax, p` (for example [here](https://github.com/MakieOrg/Makie.jl/blob/2160439bae753e54d306b6e12d305cebe5763d4e/test/statistical_tests.jl#LL41) and [here](https://github.com/MakieOrg/GraphMakie.jl/blob/85de50648d15b376434a28c6d7471c9d92f14ba0/docs/examples/plots.jl#LL140)). The `ax` is useful because it enables the user to modify the axis of the plot.
The `display` is necessary since without it you would have to do first `f, ax, p = plot(...)` and then manually `display(f)` to see the results plotted.

test/Visualization_test.jl
jofrevalles: This block of code gives error, maybe we have to look at this since it only fails when the `labels` of the tensor with a single dimension is a `tuple. For example this fails:
```julia
tn = TensorNetwork([
           Tensor(rand(2, 2, 2, 2), [:x, :y, :z, :t]),
           Tensor(rand(2, 2), [:x, :y]),
           Tensor(rand(2),(:x))
           ])
```
But this doesn't:
```julia
tn = TensorNetwork([
           Tensor(rand(2, 2, 2, 2), (:x, :y, :z, :t)),
           Tensor(rand(2, 2), (:x, :y)),
           Tensor(rand(2),[:x])
           ])
```
For now, I will keep the labels in a `Vector`.

mofeing: Ah yes, it’s because a single element tuple needs a coma so `(:x,)`

.github/workflows/CI.yml
```diff
@@ -42,8 +42,8 @@ jobs:
...
         with:
-          cache-name: CI - Julia ${{ matrix.version }} - ${{ matrix.os }} - ${{ matrix.arch }}
-          cache-compiled: true
+           cache-name: CI - Julia ${{ matrix.version }} - ${{ matrix.os }} - ${{ matrix.arch }}
+           cache-compiled: true
```
mofeing, line 46: You added some extra whitespace:
```suggestion
          cache-name: CI - Julia ${{ matrix.version }} - ${{ matrix.os }} - ${{ matrix.arch }}
          cache-compiled: true
```

mofeing (approved):

src/Visualization.jl
mofeing: Okay, I see that `f, ax, p` on `plot` functions but on `plot!` functions, only `p` is returned (currently `p` and `ax` are returned).

I'm still not sure about the `display`. `GraphMakie` does not call `display` for example.

jofrevalles: Okay, we can remove it then.

mofeing: Do you know if other `plot` functions in `Makie` are automatically displayed without calling to `display`?

jofrevalles: I tested it now and only the exclamation mark functions (e.g., `plot!`) are displayed automatically, the others don't.

jofrevalles: Sorry, I deleted the last comment since I said it the other way around: Only the exclamation mark functions (e.g., `plot!`) are not displayed automatically, the others are. So we actually do need the `display` in `plot`, it is not necessary for `plot!`. I will change it as it was.

mofeing: And how does other `Makie` projects display? Do they call `display`? We have to inspect the source code.

mofeing: Okay, this makes sense: they return `FigureAxisPlot` on `plot` and `AxisPlot` on `plot!`.

> ... We could do the same calling by `graphplot` instead of `graphplot!`, but we would need to refactor the code so we could not use exactly the same code for our `plot` and `plot!` functions. ...

Great. In order to not duplicate the code, try to abstract common code.

jofrevalles: Now the `plot` functions return the desired structs. I have combined the code into some functions, so we don't duplicate the code.

mofeing (approved):

Issue #40, opened 2023-03-24
Adjoint `Tensor` contraction produces incoherent results


jofrevalles:
### Summary
The contraction of an adjoint `Tensor` produces incoherent results with the normal `Tensor` ones. Specifically, it returns an `Array` type of data instead of `Tensor` type. I believe that is also why the PR #39 is failing on some tests.

### Example
```julia
julia> using Tenet
julia> b = Tensor(rand(2, 2), (:i, :j))
2×2 Tensor{Float64, 2, Matrix{Float64}}:
 0.604068  0.472015
 0.732286  0.0951967
julia> contract(5,b)
2×2 Tensor{Float64, 2, Matrix{Float64}}:
 3.02034  2.36008
 3.66143  0.475984
julia> contract(5,b')
2×2 Matrix{Float64}:
 3.02034  3.66143
 2.36008  0.475984
```

I believe that this is a problem with the adjoint `Tensor` not being a `Tensor` type:
```julia
julia> using Tenet
julia> b = Tensor(rand(2, 2), (:i, :j))
julia> b' isa Tensor
false
```

jofrevalles:
@mofeing Do you agree that `adjoint(a::Tensor)` should be also a `Tensor`?

mofeing:
Nice catch. Add a test for this pls.



> @mofeing Do you agree that `adjoint(a::Tensor)` should also be a `Tensor`?

I agree. But it is a little tricky in the sense that not every N-dim array has `adjoint` defined to it.

The methods should be sth like:

```julia
adjoint(t::Tensor{T,2,A}) where {T,A<:AbstractMatrix{T}} = Tensor(adjoint(parent(t)), reverse(labels(t))
adjoint(t::Tensor{T,1,A}) where {T,A<:AbstractVector{T}} = Tensor(conj(parent(t)), labels(t))
```

jofrevalles:
With the last PR with is not a problem anymore:
```julia
julia> using Tensors
julia> a = Tensor(rand(Complex{Float64}, 2, 2), (:i, :j));
julia> b = adjoint(a)
2×2 Tensor{ComplexF64, 2, Matrix{ComplexF64}}:
 0.798311-0.698864im  0.314514-0.77349im
 0.562423-0.160019im  0.986719-0.297658im
julia> b isa Tensor
true
julia> b.labels
(:i, :j)
julia> contract(a, b)
[...]

PR #39, opened 2023-03-23
Fix chain rules for contraction 

jofrevalles:
### Summary
The previous code failed the `test_rrule` and `test_frule` for complex numbers (#38). In this PR we address this issue and fix the chain `rrule` for contraction. We also added `rrule` and `frule` tests for the contraction of Tensors with complex values.

### Extra
The code now is not fully functional, and I believe that is because we have to somehow introduce the `ChainRulesCore.ProjectTo` for the adjoint `Tensor` types. I think that is the reason why `ȧ` or `ḃ` return a `Matrix` instead of a `Tensor` sometimes:
```julia
julia> @testset "[number-tensor product]" begin
           b = Tensor(rand(2, 2), (:i, :j))
           z = 1.0 + 1im

           test_frule(contract, 5.0, b)
           test_rrule(contract, 5.0, b)
       end
test_rrule: contract on Float64,Tensor{Float64, 2, Matrix{Float64}}: Error During Test at /home/jofrevalles/.julia/packages/ChainRulesTestUtils/lERVj/src/testers.jl:202
  Got exception outside of a @test
  MethodError: no method matching (::ProjectTo{Float64, NamedTuple{(), Tuple{}}})(::Matrix{Float64})
  Closest candidates are:
    (::ProjectTo)(::Thunk) at ~/.julia/packages/ChainRulesCore/a4mIA/src/projection.jl:124
    (::ProjectTo{<:Number})(::Tangent{<:Complex}) at ~/.julia/packages/ChainRulesCore/a4mIA/src/projection.jl:192
    (::ProjectTo{<:Number})(::Tangent{<:Number}) at ~/.julia/packages/ChainRulesCore/a4mIA/src/projection.jl:193
    ...
  Stacktrace:
[...]

PR #41, opened 2023-03-24
Add custom `adjoint` method for `Tensor` type

jofrevalles:
### Summary
In this PR we fix the bug found in issue #40:
- Added custom `Base.adjoint` functions for `Tensor` types. Only for two-dimensional `Tensor`s, the `labels` are `reverse`d.
- Added tests for `Base.adjoint`.

### Example
```julia
julia> using Tenet
julia> b = Tensor(rand(Complex{Float64}, 2, 2), (:i, :j))
2×2 Tensor{ComplexF64, 2, Matrix{ComplexF64}}:
 0.560191+0.926476im  0.381094+0.917558im
 0.703793+0.5878im    0.741097+0.658213im
julia> b'
2×2 Tensor{ComplexF64, 2, LinearAlgebra.Adjoint{ComplexF64, Matrix{ComplexF64}}}:
 0.560191-0.926476im  0.703793-0.5878im
 0.381094-0.917558im  0.741097-0.658213im
julia> b'.labels
(:j, :i)
```


src/Tensor.jl
```diff
@@ -151,6 +151,10 @@ function Base.view(t::Tensor, inds::Pair{Symbol,<:Any}...)
     end
 end
 
+Base.adjoint(t::Tensor{T,1,A}) where {T,A<:AbstractVector{T}} = Tensor(conj(parent(t)), labels(t))
+Base.adjoint(t::Tensor{T,2,A}) where {T,A<:AbstractMatrix{T}} = Tensor(adjoint(parent(t)), reverse(labels(t)))
+Base.adjoint(t::Tensor{T,N,A}) where {T,N,A<:AbstractArray{T, N}} = Tensor(conj(parent(t)), labels(t))
```
mofeing, line 156: Don't create this method because it doesn't even exist for N-dim Arrays.

test/Tensor_test.jl
```diff
@@ -124,6 +124,24 @@
...
+    @testset "adjoint" begin
+        using Tenet: adjoint
+
+        data = rand(2, 2, 2)
+        tensor = Tensor(data, (:i, :j, :k))
+
+        @test adjoint(tensor) |> labels == (:i, :j, :k)
+        @test adjoint(tensor) |> parent == conj(data)
+        @test adjoint(tensor).meta == tensor.meta
+
```
mofeing, line 136:
```suggestion
```

mofeing (changes requested):
Could you add a test case for a complex-vector and also nest some test set to differentiate them.

jofrevalles:
> Could you add a test case for a complex-vector and also nest some test set to differentiate them.

Okay! The test set to differentiate should be `test/Integration/ChainRules_test.jl`, right?

jofrevalles:
Something like this:
```julia
        @testset "[adjoint]" begin
            a = Tensor(rand(2, 2), (:i, :j))
            b = adjoint(a)

            test_frule(contract, a, b)
            test_rrule(contract, a, b)
        end
```

Or should I also add the `ChainRules.rrule` and `frule`s for the `adjoint` function?

mofeing:
You only need to add `rrule`s and `frule`s for `adjoint` if what you pass to an AD backend is a function **that calls `adjoint`**. That would be another issue.

This should be fine.

mofeing:
@jofrevalles tests failed.

mofeing:
Impossible because it was merged.

jofrevalles:
But this branch does not have that commit, it was created before the merge.

jofrevalles:
> > But this branch does not have that commit, it was created before the merge.
> 
> Tests run after "merging" the branches temporarily, so the repo where the tests run contain both commits from `master` and this PR. So tests run on this PR contain the commits from #39.

Okay sorry, I did not know that.
So it seems that `adjoint(Matrix)` is not a `Matrix` type, so this was causing problems. I changed this to the conjugate `conj` 
 (where `conj(Matrix) isa Matrix`) and now should be working fine.

jofrevalles:
> That's not entirely correct. It is not a `Matrix` but it is an `AbstractMatrix`.

The problem is that since it is not a `Matrix` type, somewhere it tries to convert a `Tensor` with a `Matrix` in `data` to a `Tensor` with `LinearAlgebra.Adjoint{T, Matrix{T}` data:
```julia
julia> a = Tensor(rand(2, 2), (:i, :j))
 0.813942  0.640558
 0.148885  0.77122
julia> b = adjoint(a)
2×2 Tensor{Float64, 2, LinearAlgebra.Adjoint{Float64, Matrix{Float64}}}:
 0.813942  0.148885
 0.640558  0.77122
julia> test_rrule(contract, a, b)
test_rrule: contract on Tensor{Float64, 2, Matrix{Float64}},Tensor{Float64, 2, LinearAlgebra.Adjoint{Float64, Matrix{Float64}}}: Error During Test at /home/jofrevalles/.julia/packages/ChainRulesTestUtils/lERVj/src/testers.jl:202
  Got exception outside of a @test
  MethodError: Cannot `convert` an object of type 
    Tensor{Float64{},2,Matrix{Float64}} to an object of type 
    Tensor{Float64{},2,LinearAlgebra.Adjoint{Float64, Matrix{Float64}}}
  Closest candidates are:
    convert(::Type{T}, ::LinearAlgebra.Factorization) where T<:AbstractArray at /opt/julia-1.8.2/share/julia/stdlib/v1.8/LinearAlgebra/src/factorization.jl:58
    convert(::Type{T}, ::T) where T<:AbstractArray at abstractarray.jl:16
    convert(::Type{T}, ::T) where T at Base.jl:61
    ...
```

> EDIT: notice that another test fails now.

This is because I should change the test of `adjoint` from `adjoint(data)` to `conj(data)`. With this, it doesn't fail.

Right now we are not using `adjoint` in `ChainRules`, we are using `conjugate` since it is general. However, if you prefer to keep the `adjoint` for two-dimensional tensors, we can maybe extend the `convert` function so it knows how to handle `LinearAlgebra.Adjoint{Matrix}` types.

mofeing:
Okay, let's discuss it in person.

src/Tensor.jl
```diff
@@ -151,6 +151,9 @@ function Base.view(t::Tensor, inds::Pair{Symbol,<:Any}...)
     end
 end
 
+Base.adjoint(t::Tensor{T,1,A}) where {T,A<:AbstractVector{T}} = Tensor(conj(parent(t)), labels(t))
+Base.adjoint(t::Tensor{T,2,A}) where {T,A<:AbstractMatrix{T}} = Tensor(conj(parent(t)), reverse(labels(t)))
```
mofeing, line 155: If it is just applying the conjugate, should it reverse the labels?

jofrevalles: Yes, maybe it's better not to `reverse` the `labels`. I will update that.

So do you agree it's a good idea to put `conj` instead of `adjoint` here?

jofrevalles: Yes, that's true. We can take a note (or add a comment) and consider this in the future.

mofeing:
@jofrevalles I've found one of this edge cases: matrix multiplication.

## Old definition with `adjoint`
If we define `adjoint(::Tensor)` as:
```julia
Base.adjoint(t::Tensor{T,2,A}) where {T,A<:AbstractMatrix{T}} = Tensor(adjoint(parent(t.data)), reverse(labels(t)); t.meta...)
```
then 
```julia
julia> using Tenet, LinearAlgebra

julia> Y = Tensor([0 -im; im 0], (:i, :j))
2×2 Tensor{Complex{Int64}, 2, Matrix{Complex{Int64}}}:
 0+0im  0-1im
 0+1im  0+0im

julia> Y'
2×2 Tensor{Complex{Int64}, 2, LinearAlgebra.Adjoint{Complex{Int64}, Matrix{Complex{Int64}}}}:
 0+0im  0-1im
 0+1im  0+0im

julia> Y * Y'
2×2 Matrix{Complex{Int64}}:
 1+0im  0+0im
 0+0im  1+0im

julia> tr(Y * Y')
2 + 0im

julia> contract(Y,Y')
0-dimensional Tensor{Complex{Int64}, 0, Array{Complex{Int64}, 0}}:
2 + 0im
```

## New definition with `conj`
But if we define `adjoint(::Tensor)` as...
```julia
Base.adjoint(t::Tensor{T,2,A}) where {T,A<:AbstractMatrix{T}} = Tensor(conj(parent(t.data)), labels(t); t.meta...)
```

then matrix multiplication `*` and `contract` mismatch
```julia
julia> Y
2×2 Tensor{Complex{Int64}, 2, Matrix{Complex{Int64}}}:
 0+0im  0-1im
 0+1im  0+0im

julia> Y'
2×2 Tensor{Complex{Int64}, 2, Matrix{Complex{Int64}}}:
 0+0im  0+1im
 0-1im  0+0im

julia> Y * Y'
2×2 Matrix{Complex{Int64}}:
 -1+0im   0+0im
  0+0im  -1+0im

julia> tr(Y * Y')
-2 + 0im

julia> contract(Y, Y')
0-dimensional Tensor{Complex{Int64}, 0, Array{Complex{Int64}, 0}}:
2 + 0im

julia> transpose(Y')
2×2 transpose(::Tensor{Complex{Int64}, 2, Matrix{Complex{Int64}}}) with eltype Complex{Int64}:
 0+0im  0-1im
 0+1im  0+0im

julia> tr(Y * transpose(Y'))
2 + 0im
```

## Summary
`contract` is always correct while `*` might not. I think I prefer the `adjoint` call to ensure compatibility with other Julia code right now.

There is an easy solution: instead of just calling `adjoint`, we can call `copy ∘ adjoint`. This way the `Adjoint` structure disappears and gets converted into a `Matrix` again.

```julia
julia> adjoint(parent(Y)) |> typeof
Adjoint{Complex{Int64}, Matrix{Complex{Int64}}}

julia> adjoint(parent(Y)) |> copy |> typeof
Matrix{Complex{Int64}} (alias for Array{Complex{Int64}, 2})
```

We wouldn't benefit from lazy views of the original array (i.e. `Y` is not copied and then transformed when calling `Y'`), but this is something that we also lose when calling `conj`. I'm uploading this change to check it out against the tests.

mofeing:
Btw, the contents of the metadata of the tensors were being tested... but they were initialized to an empty `Dict` so it was `true` always. 😆

If it is initialized to something, it fails. I'm submitting a fix also.

PR #54, opened 2023-04-27
Fix `rand` function for odd-length `MatrixProductState`

jofrevalles:
This PR addresses the issue with the `rand` function for generating random `MatrixProductStates` (MPS) with an odd number of tensors. Previously, the function produced an error when the number of tensors in the argument was odd:
```julia
julia> ψ = rand(MatrixProductState{Open}, 7, 2, 32)
ERROR: DimensionMismatch: size(9edf0aa5-219f-4849-99d3-a85b569b6773)=16 but should be 8
```

Now, the `rand` function correctly handles cases where the number of tensors is odd. Also, we have added tests to ensure its correct behavior.

### DIFF
--- a/src/Quantum/MatrixProductState.jl
+++ b/src/Quantum/MatrixProductState.jl
@@ -153,7 +153,7 @@ function Base.rand(rng::Random.AbstractRNG, sampler::MPSSampler{Open,T}) where {
             χr = min(χ, p^i)
 
             # swap bond dims after mid and handle midpoint for odd-length MPS
-            (n % 2 == 1 && i == n ÷ 2 + 1) ? (χl, χl) : (after_mid ? (χr, χl) : (χl, χr))
+            (isodd(n) && i == n ÷ 2 + 1) ? (χl, χl) : (after_mid ? (χr, χl) : (χl, χr))
         end
 
         # fix for first site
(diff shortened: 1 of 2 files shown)


src/Quantum/MatrixProductState.jl
```diff
@@ -151,8 +152,8 @@ function Base.rand(rng::Random.AbstractRNG, sampler::MPSSampler{Open,T}) where {
             χl = min(χ, p^(i - 1))
             χr = min(χ, p^i)
 
-            # swap bond dims after mid
-            after_mid ? (χr, χl) : (χl, χr)
+            # swap bond dims after mid and handle midpoint for odd-length MPS
+            (n % 2 == 1 && i == n ÷ 2 + 1) ? (χl, χl) : (after_mid ? (χr, χl) : (χl, χr))
```
mofeing, line 156: Use `isodd`. It is easier to read.
```suggestion
            (isodd(n) && i == n ÷ 2 + 1) ? (χl, χl) : (after_mid ? (χr, χl) : (χl, χr))
```

mofeing (changes requested):

mofeing:
Great! Thanks for catching this bug. Just a lil' change and I'm merging it!

PR #60, opened 2023-05-25
Implement `RankSimplification` transformation for `TensorNetwork`

### LINKED ISSUE #15: Implement "Rank simplification" transformation

jofrevalles:
### Summary
This PR addresses the issue #15 (resolves #15) by introducing the `RankSimplification` transformation in the `transform!` function for `TensorNetwork`s. This transformation applies the rank simplification (as defined [here](https://arxiv.org/pdf/2002.01935.pdf)) to simplify the tensor network without increasing the rank of any tensors. The key idea is to perform contractions between connected tensors that don't increase the rank of the tensors, potentially reducing the complexity of the network.

In addition to implementing the transformation, this PR includes tests to ensure the correctness and robustness of the new method.

### Example
```julia
julia> using Tenet

julia> B = Tensor(rand(2, 2), (:j, :l)) # This tensor can be absorbed by a neighbor tensor
2×2 Tensor{Float64, 2, Matrix{Float64}}: ...

julia> C = Tensor(rand(2, 2, 2, 2), (:l, :m, :o, :p))
2×2×2×2 Tensor{Float64, 4, Array{Float64, 4}}: ...

julia> tn = TensorNetwork([A, B, C])
TensorNetwork{Arbitrary}(#tensors=3, #inds=7)

julia> rank_reduced = transform(tn, Tenet.RankSimplification)
TensorNetwork{Arbitrary}(#tensors=2, #inds=6)

julia> [labels(tensor) for tensor in rank_reduced.tensors]
2-element Vector{Tuple{Symbol, Symbol, Symbol, Vararg{Symbol}}}:
 (:i, :k, :l)
 (:l, :m, :o, :p)
```
In this example, the rank simplification transformation is applied to the tensor network of three tensors. This results in a network with only two tensors where the labels and indices are updated accordingly.


mofeing:
Great work. I've just rewritten it so it is easier to read. Also, you should try not to use internal details because it will make more difficult to update all the code without bugs when the next refactoring comes.

PR #58, opened 2023-05-23
Add `DiagonalReduction` transformation for `TensorNetwork`

jofrevalles:
### Summary
This PR addresses the issue #14 (resolves #14) by introducing the `DiagonalReduction` transformation in the `transform!` function for `TensorNetwork`s. This transformation applies diagonal reduction (as defined [here](https://arxiv.org/pdf/2002.01935.pdf)) to simplify the tensor network while preserving its mathematical properties. The key idea is to minimize the number of indices by identifying tensors with a certain diagonal structure and collapsing these indices, potentially leading to computational efficiency.

In addition to the implementation of the transformation, this PR includes tests to ensure correctness and robustness of the new method.

### Example
```julia
julia> using Tenet

julia> data = [1.0 0.0; 0.0 1.0;;; 1.0 0.0; 0.0 1.0] # We can clearly see that the first and second indices are diagonal
2×2×2 Array{Float64, 3}:
[:, :, 1] =
 1.0  0.0
 0.0  1.0

[:, :, 2] =
 1.0  0.0
 0.0  1.0

julia> B = Tensor(rand(2, 2, 2), (:j, :l, :m))
2×2×2 Tensor{Float64, 3, Array{Float64, 3}}: ...

julia> tn = TensorNetwork([A, B])
TensorNetwork{Arbitrary}(#tensors=2, #inds=5)

julia> [labels(tensor) for tensor in tn.tensors]
2-element Vector{Tuple{Symbol, Symbol, Vararg{Symbol}}}:
 (:i, :k)
 (:i, :l, :m)
```
This example demonstrates how `DiagonalReduction` effectively reduces the dimension of the first tensor and updates the rest of the network accordingly.

PR #61, opened 2023-05-26
Add `AntiDiagonalGauging` transformation for `TensorNetwork`

### LINKED ISSUE #16: Implement "Anti-diagonal gauging" transformation

jofrevalles:
### Summary
This PR addresses the issue #16 (resolves #16)  by introducing the `AntiDiagonalGauging` transformation in the `transform!` function for `TensorNetworks`. This transformation applies the anti-diagonal gauging (as defined [here](https://arxiv.org/pdf/2002.01935.pdf)) to modify the tensor network while preserving its mathematical properties.

The `AntiDiagonalGauging` transformation reverses the order of some indices for tensors that fulfill an anti-diagonal condition. While this transformation alone does not directly contribute to computational efficiency, it prepares the tensor network for further simplifications by creating tensors with diagonal indices. These tensors can then be simplified using the `DiagonalReduction` transformation, which was implemented in PR #58.

In addition to the implementation of the transformation, this PR includes tests to ensure the correctness and robustness of the new method.

### Example
```julia
julia> using Tenet

julia> data = [0.0 1.0; 1.0 0.0;;; 0.0 1.0; 1.0 0.0] # The first and second indices are anti-diagonal
2×2×2 Array{Float64, 3}:
[:, :, 1] =
 0.0  1.0
 1.0  0.0

[:, :, 2] =
 0.0  1.0
 1.0  0.0

julia> A = Tensor(data, (:i, :j, :k))

julia> B = Tensor(rand(2, 2, 2), (:j, :k, :l))

TensorNetwork{Arbitrary}(#tensors=2, #inds=4)

julia> gauged_tn = transform(tn, Tenet.AntiDiagonalGauging)
TensorNetwork{Arbitrary}(#tensors=2, #inds=4)

julia> tensors(gauged_tn)[1] # We see that it now has a diagonal structure
2×2×2 Tensor{Float64, 3, Array{Float64, 3}}:
[:, :, 1] =
 1.0  0.0
 0.0  1.0

[:, :, 2] =
 1.0  0.0
 0.0  1.0

julia> contract(A, B) ≈ contract(tensors(gauged_tn)[1], tensors(gauged_tn)[2])
true
```

In this example, we illustrate how the `AntiDiagonalGauging` transformation converts a tensor with an anti-diagonal structure into one with a diagonal structure.

### DIFF
--- a/src/Transformations.jl
+++ b/src/Transformations.jl
@@ -3,7 +3,7 @@ using OMEinsum
 
 abstract type Transformation end
 
-transform(tn::TensorNetwork, transformations) = (tn = copy(tn); transform!(tn, transformations); return tn)
+transform(tn::TensorNetwork, transformations) = (tn = deepcopy(tn); transform!(tn, transformations); return tn)
 
 function transform! end
 
--- a/src/Transformations.jl
+++ b/src/Transformations.jl
@@ -114,10 +114,8 @@ end
 
 function transform!(tn::TensorNetwork, config::AntiDiagonalGauging)
     skip_inds = isempty(config.skip) ? openinds(tn) : config.skip
-    queue = collect(keys(tn.tensors))
 
-    while !isempty(queue) # loop over all tensors
-        idx = pop!(queue)
+    for idx in keys(tn.tensors)
         tensor = tn.tensors[idx]
 
         anti_diag_axes = find_anti_diag_axes(parent(tensor), config.atol)
--- a/src/Transformations.jl
+++ b/src/Transformations.jl
@@ -107,4 +107,76 @@ function find_diag_axes(x::AbstractArray, atol=1e-12)
     end
 end
 
+Base.@kwdef struct AntiDiagonalGauging <: Transformation
+    atol::Float64 = 1e-12
+    skip::Vector{Index} = Symbol[]
+end
+
+function transform!(tn::TensorNetwork, config::AntiDiagonalGauging)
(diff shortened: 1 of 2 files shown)


src/Transformations.jl
```diff
@@ -3,7 +3,7 @@ using OMEinsum
 
 abstract type Transformation end
 
-transform(tn::TensorNetwork, transformations) = (tn = copy(tn); transform!(tn, transformations); return tn)
+transform(tn::TensorNetwork, transformations) = (tn = deepcopy(tn); transform!(tn, transformations); return tn)
```
mofeing, line 6: Why is the `deepcopy` needed? Most probably is needed but I'm curious.

src/Transformations.jl
```diff
@@ -107,4 +107,50 @@ function find_diag_axes(x::AbstractArray, atol=1e-12)
...
+    atol::Float64 = 1e-12
+    skip::Vector{Index} = Symbol[]
+end
+
+function transform!(tn::TensorNetwork, config::AntiDiagonalGauging)
+    skip_inds = isempty(config.skip) ? openinds(tn) : config.skip
+    queue = collect(keys(tn.tensors))
+
+    while !isempty(queue) # loop over all tensors
+        idx = pop!(queue)
```
mofeing, line 120: Why don't you just do a `for` loop over the `keys(tn.tensors)`?

```suggestion

    for idx in keys(tn.tensors)
```

mofeing (changes requested):

PR #62, opened 2023-05-29
Fix `DiagonalReduction` tranformation for reduction over `skip_inds`

jofrevalles:
### Summary
This PR addresses a previously overlooked scenario in the `DiagonalReduction` transformation of the `transform` function. Specifically, it targets instances where the index to be reduced is included in `skip_inds`, which is set to be the open indices of the network by default. In such cases, the transformation previously failed to correctly reduce the dimension of a `Tensor` in a `TensorNetwork`.

To fix this,  we have introduced a COPY tensor whenever an index is reduced over a `Tensor` with an index in `skip_inds`. to illustrate this, let's consider a `TensorNetwork` $`TN=T^{1}_{ijk}T^{2}_{jkl} `$, and we want to reduce the diagonal indices $ij$ where $i$ is on `skip_inds` since it is an open index. A regular reduction would lead to $`\tilde{TN}=T^{1}_{ik} T^{2}_{ikl}`$, but as a result, the index $i$ would no longer be an open index, and thus $`TN \neq \tilde{TN}`$. To ensure that the dimensionality of the two expressions is the same, we need to add a COPY tensor: $` {TN}^{\prime}=T^{1}_{i_{1} k} T^{2}_{i_{2} kl} \delta_{i_{1} i_{2} i_{3}}`$.

Additionally, we have refined the corresponding tests to better accommodate this specific scenario.

### Example
Below is an example illustrating a diagonal reduction wherein one of the indices is an open index, thus introducing a COPY tensor in the reduction:
```julia
julia> using Tenet

julia> data = [1.0 0.0; 0.0 1.0;;; 1.0 0.0; 0.0 1.0] # The first and second indices are diagonal
2×2×2 Array{Float64, 3}:
[:, :, 1] =
 1.0  0.0
 0.0  1.0

[:, :, 2] =
 1.0  0.0
 0.0  1.0

julia> A = Tensor(data, (:i, :j, :k))

julia> B = Tensor(rand(2, 2, 2), (:j, :k, :l))

TensorNetwork{Arbitrary}(#tensors=2, #inds=4)

julia> reduced = transform(tn, DiagonalReduction)
TensorNetwork{Arbitrary}(#tensors=3, #inds=6)

julia> [labels(tensor) for tensor in reduced.tensors]
3-element Vector{Tuple{Symbol, Symbol, Vararg{Symbol}}}:
 (:i1, :k)
 (:i2, :k, :l)
 (:i1, :i2, :i3)
```

### DIFF
--- a/src/Transformations.jl
+++ b/src/Transformations.jl
@@ -1,5 +1,6 @@
 using DeltaArrays
 using OMEinsum
+using UUIDs: uuid4
 
 abstract type Transformation end
 
@@ -90,8 +91,8 @@ function transform!(tn::TensorNetwork, config::DiagonalReduction)
             # if the new index is in skip_inds, we need to add a COPY tensor
             if new ∈ nameof.(skip_inds)
                 data = DeltaArray{replacements+2}(ones(size(tensor, new))) # +2 for the new COPY tensor and the old index
-                indices = [Symbol("$(new)$i") for i in 1:replacements+2]
-                copy_tensor = Tensor(data, indices)
+                indices = [Symbol(uuid4()) for i in 1:replacements+1]
+                copy_tensor = Tensor(data, (new, indices...))
 
                 # replace the new index in the other tensors in the network
                 counter = 1
--- a/src/Transformations.jl
+++ b/src/Transformations.jl
@@ -65,14 +65,16 @@ function transform!(tn::TensorNetwork, config::DiagonalReduction)
             ix_i, ix_j = labels(tensor)[i], labels(tensor)[j]
 
             # do not reduce output indices
-            new, old = (ix_j in skip_inds) ? ((ix_i in skip_inds) ? continue : (ix_j, ix_i)) : (ix_i, ix_j)
+            new, old = (ix_j in nameof.(skip_inds)) ? ((ix_i in nameof.(skip_inds)) ? continue : (ix_j, ix_i)) : (ix_i, ix_j)
 
             # replace old index in the other tensors in the network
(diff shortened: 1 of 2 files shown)


mofeing:
I don't like the idea that the output index gets renamed. Could you let the output index name untouched?

For the new indices, better use `UUIDs.uuid4()`.

jofrevalles:
> I don't like the idea that the output index gets renamed. Could you let the output index name untouched?
> 
> For the new indices, better use `UUIDs.uuid4()`.

Okay, now this is fixed:
```julia
...

TensorNetwork{Arbitrary}(#tensors=2, #inds=4)

julia> [labels(tensor) for tensor in tn.tensors]
2-element Vector{Tuple{Symbol, Symbol, Symbol}}:
 (:i, :j, :k)
 (:j, :k, :l)

julia> reduced = transform(tn, DiagonalReduction)
TensorNetwork{Arbitrary}(#tensors=3, #inds=5)

julia> [labels(tensor) for tensor in reduced.tensors]
3-element Vector{Tuple{Symbol, Symbol, Vararg{Symbol}}}:
 (Symbol("18fe49cf-b246-444a-a4ce-60eb7337373d"), :k)
 (Symbol("7a09ade8-5d19-4462-b0dc-e6826b37f44f"), :k, :l)
 (:i, Symbol("18fe49cf-b246-444a-a4ce-60eb7337373d"), Symbol("7a09ade8-5d19-4462-b0dc-e6826b37f44f"))
 ```

PR #63, opened 2023-05-31
Implement `ColumnReduction` transformation for `TensorNetwork`

### LINKED ISSUE #17: Implement "Column reduction" transformation

jofrevalles:
### Summary
This PR addresses the issue #17 (resolves #17) by introducing the `ColumnReduction` transformation in the `transform!` function for `TensorNetwork`s. This transformation applies the column reduction (as defined [here](https://arxiv.org/pdf/2002.01935.pdf)) to simplify the tensor network. The key idea is to reduce the rank of the tensors that have a dimension where all but a "column" is filled with zeros.

Furthermore, we have extended the scope of this method to handle cases where more than one column contains non-zeros but specific columns are entirely zero-filled. This allows the reduction of the index dimensionality where columns are filled with zeros.

In addition to implementing the transformation, this PR includes tests to ensure the correctness and robustness of the new method.

### Example
The primary function of `ColumnReduction` is to decrease tensor rank where a dimension features only one non-zero 'column'. The following example illustrates this operation:
```julia
julia> using Tenet

julia> data = rand(3, 3, 3)
...

julia> data[:,1:2,:] .= 0 # 1st and 2nd column of the 2nd dimension are zero
...

julia> data # only 3rd columns is non-zero
3×3×3 Array{Float64, 3}:
[:, :, 1] =
 0.0  0.0  0.350011
 0.0  0.0  0.990275
 0.0  0.0  0.295394

[:, :, 2] =
 0.0  0.0  0.538989
 0.0  0.0  0.141018
 0.0  0.0  0.687099

[:, :, 3] =
 0.0  0.0  0.291134
 0.0  0.0  0.0972771
 0.0  0.0  0.614801

julia> A = Tensor(data, (:i, :j, :k)); B = Tensor(rand(3, 3), (:j, :l)); C = Tensor(rand(3, 3), (:j, :m))
3×3×3 Tensor{Float64, 3, Array{Float64, 3}}:
...

julia> tn = TensorNetwork([A, B, C])
TensorNetwork{Arbitrary}(#tensors=3, #inds=5)

julia> reduced = transform(tn, ColumnReduction)

julia> [labels(tensor) for tensor in reduced.tensors]
3-element Vector{Tuple{Symbol, Vararg{Symbol}}}:
 (:i, :k)
 (:l,)
 (:m,)
```
In this example, the column reduction transformation results in a network where the tensors are unconnected.

We now show how we enhanced the transformation to operate in scenarios where more than one non-zero column exists. The example below highlights such an instance:
```julia
julia> using Tenet

julia> data = rand(3, 3, 3)
...

julia> data[:,2,:] .= 0 # 2nd column of the 2nd dimension can be reduced
...

julia> data # 1st and 3rd column of the 2nd dimension are non-zero
3×3×3 Array{Float64, 3}:
[:, :, 1] =
 0.615815  0.0  0.245444
 0.26432   0.0  0.0888118
 0.812256  0.0  0.294028

[:, :, 2] =
 0.998973  0.0  0.0581316
 0.532386  0.0  0.764539
 0.503775  0.0  0.812196

[:, :, 3] =
 0.959305  0.0  0.812671
 0.717794  0.0  0.925235
 0.812287  0.0  0.54355

julia> A = Tensor(data, (:i, :j, :k)); B = Tensor(rand(3, 3), (:j, :l)); C = Tensor(rand(3, 3), (:j, :m))
3×3×3 Tensor{Float64, 3, Array{Float64, 3}}:
...

julia> tn = TensorNetwork([A, B, C])
TensorNetwork{Arbitrary}(#tensors=3, #inds=5)

julia> reduced = transform(tn, ColumnReduction)
TensorNetwork{Arbitrary}(#tensors=3, #inds=5)

julia> [size(tensor) for tensor in reduced.tensors]
3-element Vector{Tuple{Int64, Int64, Vararg{Int64}}}:
 (3, 2, 3)
 (2, 3)
 (2, 3)
```
In this example we show that when we have only one zero column in an index, we can apply a transformation to reduce the dimension of that index.

### DIFF
--- a/src/Transformations.jl
+++ b/src/Transformations.jl
@@ -243,18 +243,26 @@ function transform!(tn::TensorNetwork, config::ColumnReduction)
 end
 
 function find_zero_columns(x; atol=1e-12)
-    ndims = size(x)
+    dims = size(x)
 
-    column_axes = []
-    for d in 1:length(ndims)
-        for i in 1:ndims[d]
-            column = selectdim(x, d, i)
-            if all(abs.(column) .<= atol)
-                push!(column_axes, (d, i))
+    # Create an initial set of all possible column pairs
+    zero_columns = Set((d, c) for d in 1:length(dims) for c in 1:dims[d])
+
+    # Iterate over each element in tensor
+    for index in CartesianIndices(x)
+        val = x[index]
+
+        # For each non-zero element, eliminate the corresponding column from the zero_columns set
+        if abs(val) > atol
+            for d in 1:length(dims)
+                c = index[d]
+                delete!(zero_columns, (d, c))
             end
         end
     end
-    return unique(column_axes)
+
+    # Now the zero_columns set only contains column pairs where all elements are zero
+    return collect(zero_columns)
 end
 
 function find_diag_axes(x; atol = 1e-12)
--- a/src/Transformations.jl
+++ b/src/Transformations.jl
@@ -217,6 +217,7 @@ function transform!(tn::TensorNetwork, config::ColumnReduction)
                     tn.tensors[ind] = Tensor(new_tensor, new_labels)
                 end
             end
+            delete!(tn.inds, ix_i)
         end
 
         # Then try to reduce the dimensionality of the index in the other tensors
@@ -236,7 +237,6 @@ function transform!(tn::TensorNetwork, config::ColumnReduction)
                     tn.tensors[ind] = Tensor(view(parent(t), reduced_dims...), labels(t))
                 end
             end
-            delete!(tn.inds, ix_i)
         end
     end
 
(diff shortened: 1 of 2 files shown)


src/Transformations.jl
```diff
@@ -178,6 +178,86 @@ function transform!(tn::TensorNetwork, config::AntiDiagonalGauging)
...
+Base.@kwdef struct ColumnReduction <: Transformation
+    atol::Float64 = 1e-12
+    skip::Vector{Index} = Symbol[]
+end
+
+function transform!(tn::TensorNetwork, config::ColumnReduction)
+    skip_inds = isempty(config.skip) ? openinds(tn) : config.skip
+
+    for idx in keys(tn.tensors)
+        tensor = tn.tensors[idx]
```
mofeing, line 190: Why not just iterate over `tn.tensors`? `idx` is not used anywhere else.
```suggestion
    for tensor in tn.tensors
```

jofrevalles: Yes, I missed that!

mofeing (approved):

PR #70, opened 2023-07-03
Implement `SplitSimplification` transformation for `TensorNetwork`

### LINKED ISSUE #18: Implement "Split simplification" transformation

jofrevalles:
### Summary
This PR introduces the `SplitSimplification` transformation as part of the `transform!` function for `TensorNetwork`s, addressing issue #18 (resolve #18). The transformation applies the split simplification procedure, as outlined in [this paper](https://arxiv.org/pdf/2002.01935.pdf), to reduce the complexity of the tensor network while ensuring the rank of any tensor does not increase. The main strategy involves performing a singular value decomposition across all possible bipartitions of a tensor's indices, and replacing the tensor with a lower-rank approximation when this results in a simplification of the `TensorNetwork`.

Additionally, this PR includes comprehensive tests that verify the correctness and robustness of the newly implemented transformation.

### Example
```julia
julia> using Tenet

julia> v1 = Tensor([1, 2, 3], (:i,))
3-element Tensor{Int64, 1, Vector{Int64}}:
...

julia> v2 = Tensor([4, 5, 6], (:j, ))
3-element Tensor{Int64, 1, Vector{Int64}}:
...

julia> m1 = Tensor(rand(3, 3), (:k, :l))
3×3 Tensor{Float64, 2, Matrix{Float64}}:
...

julia> t1 = contract(v1, v2)
3×3 Tensor{Int64, 2, Matrix{Int64}}:
...

julia> tensor = contract(t1, m1) # Define a tensor which can be splitted in three
3×3×3×3 Tensor{Float64, 4, Array{Float64, 4}}:
...

julia> tn = TensorNetwork([tensor, Tensor(rand(3, 3, 3), (:k, :m, :n)), Tensor(rand(3, 3, 3), (:l, :n, :o))])
TensorNetwork{Arbitrary, NamedTuple{(), Tuple{}}}(#tensors=3, #labels=7)

julia> reduced = transform(tn, SplitSimplification)
TensorNetwork{Arbitrary, NamedTuple{(), Tuple{}}}(#tensors=5, #labels=9)

julia> size.(tensors(tn))
3-element Vector{Tuple{Int64, Int64, Int64, Vararg{Int64}}}:
 (3, 3, 3, 3)
 (3, 3, 3)
 (3, 3, 3)

julia> size.(tensors(reduced))
5-element Vector{Tuple{Int64, Int64, Vararg{Int64}}}:
 (3, 3, 3)
 (3, 3, 3)
 (3, 1)
 (3, 1)
 (1, 1, 3, 3)
```
In this example, the `SplitSimplification` transformation is applied to a tensor network that includes a rank-4 tensor that can be simplified using svd into two vectors and a matrix. The result is a network containing five tensors, each of lower rank and dimensions.


mofeing:
Super!

> julia> size.(tensors(reduced))
> 5-element Vector{Tuple{Int64, Int64, Vararg{Int64}}}:
>  (3, 3, 3)
>  (3, 3, 3)
>  (3, 1)
>  (3, 1)
>  (1, 1, 3, 3)
> ```

I think it's best to directly remove the indices whose resulting dimension is 1. What do you think @jofrevalles?

jofrevalles:
> I think it's best to directly remove the indices whose resulting dimension is 1. What do you think @jofrevalles?

Maybe we could extend `dropdims` for `Tensor`s, right? This would be easier.

src/Transformations.jl
```diff
@@ -260,6 +261,64 @@ function transform!(tn::TensorNetwork, config::ColumnReduction)
...
+
+                # Get the singular values and decide the rank
+                singular_values = diag(s)
+                rank_s = sum(singular_values .> config.atol)
+
+                if rank_s < length(singular_values)
+                    # Remove unnecessary data in u, s, v
+                    u = view(u, labels(s)[1] => 1:rank_s)
+                    s = view(s, (idx -> idx => 1:rank_s).(labels(s))...)
+                    v = view(v, labels(s)[2] => 1:rank_s)
```
mofeing, line 298: Maybe is better not to get views? If we get views, the underlying array will be around and it will be bigger than this views.

Also,
```suggestion
                if rank_s < size(s,1)
                    # truncate data
                    u = view(u, labels(s)[1] => 1:rank_s)
                    s = view(s, (idx -> idx => 1:rank_s).(labels(s))...)
                    v = view(v, labels(s)[2] => 1:rank_s)
```

src/Transformations.jl
```diff
@@ -260,6 +261,64 @@ function transform!(tn::TensorNetwork, config::ColumnReduction)
...
+function transform!(tn::TensorNetwork, config::SplitSimplification)
+    done = false  # This flag will be set to true when no further changes are made
+
+    while !done
+        done = true  # Assume no changes will be made this pass
+
+        # Create a copy of the list of tensors so we can modify the original while iterating
+        tensor_list = copy(tensors(tn))
+
+        for tensor in tensor_list
```
mofeing, line 277: I would prefer if you do something with `@label` and `@goto` like in `RankSimplification`.

mofeing (changes requested):

mofeing:
> > I think it's best to directly remove the indices whose resulting dimension is 1. What do you think @jofrevalles?
> 
> Maybe we could extend `dropdims` for `Tensor`s, right? This would be easier.

Yeah, that seems correct to me.

mofeing:
@jofrevalles I've implemented `dropdims` for `Tensor` and released in the new `Tensors` v0.1.9 release. Update the `compat` section to get the `dropdims` method.

src/Transformations.jl
jofrevalles: Okay, I will take a look into `RankSimplification`

mofeing:
Both conditions should be equivalent mathematically, so our implementation should be correct.

src/Transformations.jl
```diff
@@ -260,6 +261,47 @@ function transform!(tn::TensorNetwork, config::ColumnReduction)
...
+                s = view(s, (idx -> idx => 1:rank_s).(labels(s))...)
+                v = view(v, labels(s)[2] => 1:rank_s)
+
+                # Replace the original tensor with the two new ones
+                tensor_l = u * s
+                tensor_r = v
+
+                pop!(tn, tensor)  # Remove the old tensor
+                push!(tn, dropdims(tensor_l, dims = tuple(findall(size(tensor_l) .== 1)...))) # Add the new tensors
+                push!(tn, dropdims(tensor_r, dims = tuple(findall(size(tensor_r) .== 1)...)))
```
mofeing, line 295:
```suggestion
                push!(tn, dropdims(tensor_l))
                push!(tn, dropdims(tensor_r))
                pop!(tn, tensor)
```

src/Transformations.jl
```diff
@@ -260,6 +261,47 @@ function transform!(tn::TensorNetwork, config::ColumnReduction)
...
+            u, s, v = svd(tensor; left_inds=left_inds)
+            rank_s = sum(diag(s) .> config.atol)
+
+            if rank_s < size(s,1)
+                # truncate data
+                u = view(u, labels(s)[1] => 1:rank_s)
+                s = view(s, (idx -> idx => 1:rank_s).(labels(s))...)
+                v = view(v, labels(s)[2] => 1:rank_s)
+
+                # Replace the original tensor with the two new ones
```
mofeing, line 289:
```suggestion
                # replace the original tensor with factorization
```

src/Transformations.jl
```diff
@@ -260,6 +261,47 @@ function transform!(tn::TensorNetwork, config::ColumnReduction)
...
+                tensor_l = u * s
+                tensor_r = v
+
+                pop!(tn, tensor)  # Remove the old tensor
+                push!(tn, dropdims(tensor_l, dims = tuple(findall(size(tensor_l) .== 1)...))) # Add the new tensors
+                push!(tn, dropdims(tensor_r, dims = tuple(findall(size(tensor_r) .== 1)...)))
+
+                # Break the loop since we modified the network and need to recheck connections
```
mofeing, line 297:
```suggestion
                # iterator is no longer valid, so restart loop
```

mofeing (changes requested):
I'm committing this minor fixes and it should be ready to merge.

PR #72, opened 2023-07-12
Fix `DimensionMismatch` not thrown in `TensorNetwork` constructor

### LINKED ISSUE #69: `DimensionMismatch` error not raised for inconsistent dimensions in `TensorNetwork`
### Summary
In the current implementation of the `TensorNetwork` function, inconsistent dimensions of the same tensor index are not correctly checked and handled. This issue manifests when two tensors with the same index label, but have different dimensions in the same `TensorNetwork`.

### Example
```julia
julia> using Tenet

julia> A = Tensor(rand(3, 3), (:i, :j))
3×3 Tensor{Float64, 2, Matrix{Float64}}:
 0.397522  0.872019   0.541417
 0.463644  0.0310421  0.479054
 0.429812  0.923011   0.628436

julia> B = Tensor(rand(2, 2), (:j, :k))
 0.383957  0.154286
 0.454446  0.608856

julia> tn = TensorNetwork([A, B]) # No error is raised here
TensorNetwork{Arbitrary, NamedTuple{(), Tuple{}}}(#tensors=2, #labels=3)
```
In the above example, despite `:j` index having dimensions of 3 in `Tensor` `A` and 2 in `Tensor` `B`, no `DimensionMismatch` error is raised, which is counter-intuitive and can potentially lead to incorrect results.

Interestingly, the `DimensionMismatch` error is correctly thrown when using the `push!` function:
```julia
julia> push!(tn, Tensor(rand(3, 3), (:k, :l))) # Raises DimensionMismatch error
ERROR: DimensionMismatch: size(tensor,k)=3 but should be equal to size(tn,k)=2
Stacktrace:
 [1] push!(tn::TensorNetwork{Arbitrary, NamedTuple{(), Tuple{}}}, tensor::Tensor{Float64, 2, Matrix{Float64}})
   @ Tenet ~/git/Tenet.jl/src/TensorNetwork.jl:167
 [2] top-level scope
   @ REPL[22]:1
```
This suggests that the error handling mechanism might be correctly implemented in `push!` but is not properly functioning in the `TensorNetwork` constructor function.

jofrevalles:
### Summary
This PR fixes issue #69 (resolves #69). The previous implementation did not raise a `DimensionMismatch` error in the `TensorNetwork` constructor function for inconsistent dimensions across tensors sharing the same index label. This issue could lead to incorrect results when working with `TensorNetwork` objects.

In response, we have implemented a dimensionality check in the `TensorNetwork` constructor function, ensuring consistency across tensors that share an index label. This is similar to the check that already exists in the `push!` method.

Furthermore, to confirm that both the `TensorNetwork` constructor and `push!` methods are functioning correctly, we have added tests to ensure they throw a `DimensionMismatch` error when encountering inconsistent dimensions.

### Example
With our fix, a `DimensionMismatch `error is now correctly thrown:
```julia
julia> using Tenet

julia> A = Tensor(rand(3, 3), (:i, :j))
3×3 Tensor{Float64, 2, Matrix{Float64}}:
 0.0256405  0.746125  0.0352789
 0.932879   0.458685  0.726907
 0.129121   0.645359  0.941556

julia> B = Tensor(rand(2, 2), (:j, :k))
 0.0137305  0.132258
 0.671962   0.411074

julia> tn = TensorNetwork([A, B]) # Now an error is raised here
ERROR: DimensionMismatch: Tensors with index j do not all have the same size
Stacktrace:
...
```

### DIFF
--- a/src/TensorNetwork.jl
+++ b/src/TensorNetwork.jl
@@ -40,10 +40,8 @@ struct TensorNetwork{A<:Ansatz,M<:NamedTuple}
 
         # Check for inconsistent dimensions
         for (index, idxs) in indices
-            sizes = [size(tensors[i], index) for i in idxs]
-            if !all(x -> x == sizes[1], sizes)
+            allequal(i -> size(tensors[i], index), idxs) ||
                 throw(DimensionMismatch("Different sizes specified for index $index"))
-            end
         end
 
         M = Tenet.metadata(A)
--- a/src/TensorNetwork.jl
+++ b/src/TensorNetwork.jl
@@ -38,6 +38,14 @@ struct TensorNetwork{A<:Ansatz,M<:NamedTuple}
             mergewith(vcat, dict, Dict([index => [i] for index in labels(tensor)]))
         end
 
+        # Check for inconsistent dimensions
+        for (index, idxs) in indices
+            sizes = [size(tensors[i], index) for i in idxs]
+            if !all(x -> x == sizes[1], sizes)
+                throw(DimensionMismatch("Different sizes specified for index $index"))
+            end
+        end
+
         M = Tenet.metadata(A)
         metadata = M((; metadata...))
 
--- a/src/TensorNetwork.jl
+++ b/src/TensorNetwork.jl
@@ -40,7 +40,7 @@ struct TensorNetwork{A<:Ansatz,M<:NamedTuple}
 
         # Check for inconsistent dimensions
         for (index, idxs) in indices
-            allequal(i -> size(tensors[i], index), idxs) ||
+            allequal(Iterators.map(i -> size(tensors[i], index), idxs)) ||
                 throw(DimensionMismatch("Different sizes specified for index $index"))
         end
 
--- a/test/Quantum_test.jl
+++ b/test/Quantum_test.jl
@@ -16,7 +16,7 @@
     )
 
     operator = TensorNetwork{MockOperator}(
-        [Tensor(rand(2, 4, 2), (:a, :c, :d)), Tensor(rand(3, 4, 2, 5), (:b, :c, :e, :f))];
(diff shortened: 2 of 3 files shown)


jofrevalles:
It seems that the new thrown error now shows us problems in the `hcat` function

src/TensorNetwork.jl
```diff
@@ -38,6 +38,14 @@ struct TensorNetwork{A<:Ansatz,M<:NamedTuple}
             mergewith(vcat, dict, Dict([index => [i] for index in labels(tensor)]))
         end
 
+        # Check for inconsistent dimensions
+        for (index, idxs) in indices
+            sizes = [size(tensors[i], index) for i in idxs]
+            if !all(x -> x == sizes[1], sizes)
+                throw(DimensionMismatch("Different sizes specified for index $index"))
+            end
```
mofeing, line 46:
```suggestion
            allequal(i -> size(tensors[i], index), idxs) ||
                throw(DimensionMismatch("Different sizes specified for index $index"))
```

mofeing:
I've refactored some code and now it should be ok.

src/TensorNetwork.jl
```diff
@@ -38,6 +38,12 @@ struct TensorNetwork{A<:Ansatz,M<:NamedTuple}
             mergewith(vcat, dict, Dict([index => [i] for index in labels(tensor)]))
         end
 
+        # Check for inconsistent dimensions
+        for (index, idxs) in indices
+            allequal(i -> size(tensors[i], index), idxs) ||
```
mofeing, line 43:
```suggestion
            allequal(Iterators.map(i -> size(tensors[i], index), idxs)) ||
```

mofeing:
> It seems that the new thrown error now shows us problems in the `hcat` function

The code is actually ok. The problem was that `hcat` tests where connecting indices with wrong sizes, so it detected a bug!

PR #73, opened 2023-07-13
Extended documentation for `transformations.jl`

jofrevalles:
### Summary
This PR aims to enhance the documentation for the `transformations.jl` module, providing more detailed descriptions of each available transformation and including a visual example to demonstrate the impact of the transformations on a tensor network.


docs/src/transformations.md
```diff
@@ -1,14 +1,54 @@
 # Transformations
 
+In tensor network computations, it is common to apply various transformations to simplify the network structure, reduce computational cost, or prepare the network for further operations. These transformations can modify the network's structure in a variety of ways such as reducing dimensions or ranks of tensors, converting certain indices, or even reordering tensor indices.
+
+In Tenet, we provide a set of predefined transformations which you can apply to your `TensorNetwork` using both the `transform` and `transform!` functions. The `transform` function is non-mutating, meaning it will return a new `TensorNetwork` where the transformation has been applied, leaving the original network unchanged. For in-place transformation, you can use the mutating version, `transform!`.
```
mofeing, line 5:
```suggestion
In Tenet, we provide a set of predefined transformations which you can apply to your `TensorNetwork` using both the `transform`/`transform!` functions. The `transform` function is non-mutating, meaning it will return a new `TensorNetwork` where the transformation has been applied, leaving the original network unchanged. For in-place transformation, you can use the mutating version, `transform!`.
```

mofeing, line 5: This already known by using the `!` suffix.

```suggestion
In Tenet, we provide a set of predefined transformations which you can apply to your `TensorNetwork` using both the `transform` and `transform!` functions.
```

docs/src/transformations.md
```diff
@@ -1,14 +1,54 @@
...
+julia> p2 = plot!(ax2, tn2; node_size=5.)
+julia> ax1.titlesize=20
+julia> ax2.titlesize=20
+```
+```@raw html
+<figure>
+<img width=500 src="../assets/transformation.svg" alt="Before and after transformation in a Tensor Network">
+</figure>
+```
+
```
mofeing, line 30: I prefer to put examples inside the docstrings of the different `Transformation`s, so the user can check their effect.

(Use small examples)

docs/src/transformations.md
```diff
@@ -1,14 +1,54 @@
...
 Tenet.AntiDiagonalGauging
+```
+
+```@docs
 Tenet.ColumnReduction
 ```
+
+```@docs
+Tenet.SplitSimplification
+```
```
mofeing, line 54: Why this separation?
```suggestion
Tenet.DiagonalReduction
Tenet.RankSimplification
Tenet.AntiDiagonalGauging
Tenet.ColumnReduction
Tenet.SplitSimplification
```

src/Transformations.jl
```diff
@@ -261,6 +269,15 @@ function transform!(tn::TensorNetwork, config::ColumnReduction)
     return tn
 end
 
+"""
+    SplitSimplification <: Transformation
+
+Reduces the rank of some tensors in the `TensorNetwork` by splitting some tensors into two. It uses teh singular value decomposition to identify the tensors that can be split without increasing the maximum rank of the network.
```
mofeing, line 275:
```suggestion
Reduces the rank of some tensors in the `TensorNetwork` by splitting some tensors into two. It uses the singular value decomposition to identify the tensors that can be split without increasing the maximum rank of the network.
```

src/Transformations.jl
```diff
@@ -71,6 +71,8 @@ end
 """
     DiagonalReduction <: Transformation
 
+Reduces the dimension of a `Tensor` in a `TensorNetwork` when it has a pair of indices that fulfil a diagonal structure.
```
mofeing, line 74:
```suggestion
Reduce the dimension of a `Tensor` in a `TensorNetwork` when it has a pair of indices that fulfil a diagonal structure.
```

src/Transformations.jl
```diff
@@ -119,7 +121,8 @@ end
 """
     RankSimplification <: Transformation
 
-Contract tensors preemptively.
+Reduces the rank of the `TensorNetwork` by contracting some tensors preemptively.
```
mofeing, line 124:
```suggestion
Reduce the rank of the `TensorNetwork` by contracting some tensors preemptively.
```

src/Transformations.jl
```diff
@@ -155,10 +158,13 @@ end
 """
     AntiDiagonalGauging <: Transformation
 
+Modifies the `TensorNetwork` by reversing the order of some indices of tensors that fulfill an anti-diagonal condition.
```
mofeing, line 161:
```suggestion
Modify the `TensorNetwork` by reversing the order of some indices of tensors that fulfill an anti-diagonal condition.
```

src/Transformations.jl
```diff
@@ -155,10 +158,13 @@ end
...
+While this transformation doesn't directly enhance computational efficiency, it sets up the `TensorNetwork` for other operations that do.
+
 # Keyword Arguments
 
   - `atol` Absolute tolerance. Defaults to `1e-12`.
-  - `skip`
+  - `skip` A list of indices to skip. Defaults to `[]`.
```
mofeing, line 167:
```suggestion
  - `skip` List of indices to skip. Defaults to `[]`.
```

src/Transformations.jl
```diff
@@ -192,10 +198,12 @@ end
 """
     ColumnReduction <: Transformation
 
+Reduces the dimension of a `Tensor` in a `TensorNetwork` when it contains columns with all elements smaller than `atol`.
```
mofeing, line 201:
```suggestion
Reduce the dimension of a `Tensor` in a `TensorNetwork` when it contains columns with all elements smaller than `atol`.
```

src/Transformations.jl
```diff
@@ -192,10 +198,12 @@ end
...
+
 # Keyword Arguments
 
   - `atol` Absolute tolerance.
-  - `skip`
+  - `skip` A list of indices to skip. Defaults to `[]`.
```
mofeing, line 206:
```suggestion
  - `skip` List of indices to skip. Defaults to `[]`.
```

src/Transformations.jl
```diff
@@ -192,10 +198,12 @@ end
 """
     ColumnReduction <: Transformation
 
+Reduces the dimension of a `Tensor` in a `TensorNetwork` when it contains columns with all elements smaller than `atol`.
+
 # Keyword Arguments
 
   - `atol` Absolute tolerance.
```
mofeing, line 205:
```suggestion
  - `atol` Absolute tolerance. Defaults to `1e-12`.
```

src/Transformations.jl
mofeing, line 275: > ... by splitting some tensors into two.
Can you rephrase this?

```suggestion
Reduce the rank of some tensors in the `TensorNetwork` by splitting some tensors into two. It uses teh singular value decomposition to identify the tensors that can be split without increasing the maximum rank of the network.
```

src/Transformations.jl
```diff
@@ -261,6 +269,15 @@ function transform!(tn::TensorNetwork, config::ColumnReduction)
...
 end
 
+"""
+    SplitSimplification <: Transformation
+
+Reduces the rank of some tensors in the `TensorNetwork` by splitting some tensors into two. It uses teh singular value decomposition to identify the tensors that can be split without increasing the maximum rank of the network.
+
+# Keyword Arguments
+
+  - `atol` Absolute tolerance.
```
mofeing, line 279:
```suggestion
  - `atol` Absolute tolerance. Defaults to `1e-10`.
```

mofeing (changes requested):
Great work! Just a couple of comments:

- It's best practice to use the infinitive for explaining functions in the documentation (e.g. use "Reduce" instead of "Reduces". Imagine that function are like _verbs_ or _actions_).
- One of the main reasons why these methods were developed is because it drastically reduces the problem size of **contraction path search**. We should mention it.
- Reducing the maximum rank of the Tensor Network is not what it solves, but **reducing the rank (or better, size) of the involved tensors**.
- It would a good idea to cite the paper by Johnnie Gray and maybe say that we have taken inspiration? from `quimb`.
- I think it would be great to put small examples inside the docstrings for each `Transformation`, so users can really appreciate each of them.

docs/src/transformations.md
jofrevalles: Should we show a visual representation for each transformation? Maybe this would be too much

mofeing: I think this would be a good idea, but using very small tensor networks. Similar to what they do in the paper with the diagrams.

docs/src/transformations.md
mofeing: @jofrevalles what about removing these 2 phrases?

docs/src/transformations.md
mofeing: I can help with this if you need help.

jofrevalles: I will try to create some examples, I will let you know!

src/Transformations.jl
```diff
@@ -119,7 +121,8 @@ end
 """
     RankSimplification <: Transformation
 
-Contract tensors preemptively.
+Reduce the rank of the `TensorNetwork` by contracting some tensors preemptively.
```
mofeing, line 124: This doesn't reduce the rank of the `TensorNetwork`...

transformation.svg
mofeing, line 1: Are you adding this to the root of the project?

jofrevalles: I am adding this to `docs/src/assets`

mofeing: The path doesn't suggest so...

src/Transformations.jl
```diff
@@ -261,6 +269,19 @@ function transform!(tn::TensorNetwork, config::ColumnReduction)
...
 end
 
+"""
+    SplitSimplification <: Transformation
+
+Reduce the rank of some tensors in the [`TensorNetwork`](@ref) by decomposing some tensors using the singular value decomposition (SVD).
+The tensors chosen are the ones that can be decomposed without increasing the maximum rank of the network.
+
+SplitSimplification <: Transformation
+
```
mofeing, line 279: Looks like an artifact right?
```suggestion
```

src/Transformations.jl
```diff
@@ -261,6 +269,19 @@ function transform!(tn::TensorNetwork, config::ColumnReduction)
     return tn
 end
 
+"""
+    SplitSimplification <: Transformation
+
+Reduce the rank of some tensors in the [`TensorNetwork`](@ref) by decomposing some tensors using the singular value decomposition (SVD).
+The tensors chosen are the ones that can be decomposed without increasing the maximum rank of the network.
```
mofeing, line 276:
```suggestion
Reduce the rank of tensors in the [`TensorNetwork`](@ref) by decomposing them using the Singular Value Decomposition (SVD). Tensors whose factorization do not increase the maximum rank of the network are left decomposed.
```

src/Transformations.jl
```diff
@@ -192,10 +198,12 @@ end
 """
     ColumnReduction <: Transformation
 
+Reduce the dimension of a `Tensor` in a [`TensorNetwork`](@ref) when it contains columns with all elements smaller than `atol`.
```
mofeing, line 201:
```suggestion
Truncate the dimension of a `Tensor` in a [`TensorNetwork`](@ref) when it contains columns with all elements smaller than `atol`.
```

src/Transformations.jl
```diff
@@ -155,10 +158,13 @@ end
 """
     AntiDiagonalGauging <: Transformation
 
+Modify the [`TensorNetwork`](@ref) by reversing the order of some indices of tensors that fulfill an anti-diagonal condition.
```
mofeing, line 161:
```suggestion
Reverse the order of tensor indices that fulfill the anti-diagonal condition.
```

src/Transformations.jl
```diff
@@ -119,7 +121,8 @@ end
 """
     RankSimplification <: Transformation
 
-Contract tensors preemptively.
+Reduce the complexity of the [`TensorNetwork`](@ref) by contracting some tensors preemptively.
+This is done by identifying pairs of tensors in the network which can be contracted together without increasing the maximum rank of the network.
```
mofeing, line 125:
```suggestion
Preemptively contract tensors whose result doesn't increase in size.
```

docs/src/transformations.md
```diff
@@ -1,14 +1,45 @@
...
+julia> p1 = plot!(ax1, tn; node_size=5.)
+julia> ax2 = Axis(fig[1, 2], title="Transformed TensorNetwork")
+julia> p2 = plot!(ax2, tn2; node_size=5.)
+julia> ax1.titlesize=20
+julia> ax2.titlesize=20
+```
+```@raw html
+<figure>
+<img width=500 src="../assets/transformation.svg" alt="Before and after transformation in a Tensor Network">
+</figure>
```
mofeing, line 33: Replace this with a `@example` block so the plot is generated during the documentation generation process.
```suggestion
```@example
tn = TensorNetwork(...)
transformed_tn = transform(tn, Tenet.RankSimplification)

fig = Figure() # hide
ax1 = Axis(fig[1, 1]; title="Original TensorNetwork") # hide
p1 = plot!(ax1, tn; node_size=5.) # hide
ax2 = Axis(fig[1, 2], title="Transformed TensorNetwork") # hide
p2 = plot!(ax2, tn2; node_size=5.) # hide
ax1.titlesize=20 # hide
ax2.titlesize=20 # hide
fig # hide
```

docs/src/transformations.md
```diff
@@ -1,14 +1,45 @@
 # Transformations
 
+In tensor network computations, it is common to apply various transformations to simplify the network structure, reduce computational cost, or prepare the network for further operations. These transformations can modify the network's structure in a variety of ways such as reducing dimensions or ranks of tensors, converting certain indices, or even reordering tensor indices.
```
mofeing, line 3: Actually is not common, but its a good idea.
```suggestion
In tensor network computations, it is good practice to apply various transformations to simplify the network structure, reduce computational cost, or prepare the network for further operations. These transformations modify the network's structure locally by permuting, contracting, factoring or truncating tensors.
```

mofeing (changes requested):
@jofrevalles please, revise these requests.

src/Transformations.jl
jofrevalles: I think so, thanks!

docs/src/transformations.md
jofrevalles: But then I would need to load the qasm circuit using `QuacIO`

jofrevalles:
Now the docs work as expected

mofeing (approved):

mofeing:
I prefer multiple smaller examples and leave the RQC for the examples, but it's good enough right now. I will change it in the future.

PR #78, opened 2023-08-01
Fix `tensors` function for `Operator` type

jofrevalles:
### Summary
This PR addresses a bug in the `tensors` function which caused a `MethodError` when called with a `TensorNetwork` of `Operator` type. The root cause was that the function signature was designed exclusively for the `State` type. The solution was to extend the function signature to handle both `Operator` and `State` types.

In addition to the fix, this PR introduces new tests designed to cover this scenario and prevent regressions.

### Example
Here we can see an example of the previous error:
```julia
julia> _arrays = [rand(1, 1, 2, 2), rand(1, 1, 2, 2), rand(1, 1, 2, 2)]
3-element Vector{Array{Float64, 4}}:
 [0.5446558537456561;;; 0.43215020981519403;;;; 0.032683161122349036;;; 0.6659303187609796]
 [0.08054317169533542;;; 0.48062834256926257;;;; 0.28742074946598595;;; 0.17849508460251406]
 [0.7073563711310278;;; 0.12764406681391194;;;; 0.34387035001349886;;; 0.8791235663956318]

julia> ψ = MatrixProduct{Operator,Periodic}(_arrays, order = (:l, :r, :i, :o))
TensorNetwork{MatrixProduct{Operator, Periodic}, NamedTuple{(:plug, :interlayer, :χ), Tuple{Type{<:Plug}, Vector{Bijections.Bijection{Int64, Symbol}}, Union{Nothing, Int64}}}}(#tensors=3, #labels=9)

julia> tensors(ψ, 1)
ERROR: MethodError: no method matching tensors(::Type{Operator}, ::TensorNetwork{MatrixProduct{Operator, Periodic}, NamedTuple{(:plug, :interlayer, :χ), Tuple{Type{<:Plug}, Vector{Bijections.Bijection{Int64, Symbol}}, Union{Nothing, Int64}}}}, ::Int64)

Closest candidates are:
  tensors(::Type{Operator}, ::TensorNetwork{<:Quantum}, ::Any, ::Symbol)
   @ Tenet ~/.julia/packages/ValSplit/MMCz3/src/ValSplit.jl:141
  tensors(::Type{State}, ::TensorNetwork{<:Quantum}, ::Any)
   @ Tenet ~/git/Tenet.jl/src/Quantum/Quantum.jl:84
  tensors(::TensorNetwork{<:Quantum}, ::Integer, ::Any...)
   @ Tenet ~/git/Tenet.jl/src/Quantum/Quantum.jl:83

Stacktrace:
 [1] tensors(::TensorNetwork{MatrixProduct{Operator, Periodic}, NamedTuple{(:plug, :interlayer, :χ), Tuple{Type{<:Plug}, Vector{Bijections.Bijection{Int64, Symbol}}, Union{Nothing, Int64}}}}, ::Int64)
   @ Tenet ~/git/Tenet.jl/src/Quantum/Quantum.jl:83
 [2] top-level scope
   @ REPL[9]:1
```

### DIFF
--- a/src/Quantum/Quantum.jl
+++ b/src/Quantum/Quantum.jl
@@ -81,7 +81,7 @@ Return the `Tensor` connected to the [`TensorNetwork`](@ref) on `site`.
 See also: [`sites`](@ref).
 """
 tensors(tn::TensorNetwork{<:Quantum}, site::Integer, args...) = tensors(plug(tn), tn, site, args...)
-tensors(::Type{State}, tn::TensorNetwork{<:Quantum}, site) = select(tn, labels(tn, :plug, site)) |> only
+tensors(::Union{Type{Operator}, Type{State}}, tn::TensorNetwork{<:Quantum}, site) = select(tn, labels(tn, :plug, site)) |> only
 @valsplit 4 tensors(T::Type{Operator}, tn::TensorNetwork{<:Quantum}, site, dir::Symbol) =
     throw(MethodError(sites, "dir=$dir not recognized"))
 
--- a/test/MatrixProductOperator_test.jl
+++ b/test/MatrixProductOperator_test.jl
@@ -76,6 +76,14 @@
                           only(select(ψ, last(ψ.interlayer)[3])).meta[:alias][:l]
                 end
             end
+
+            @testset "tensors" begin
+                arrays = [rand(1, 2, 2), rand(1, 1, 2, 2), rand(1, 2, 2)]
+                ψ = MatrixProduct{Operator,Open}(arrays, order = (:l, :r, :i, :o))
+
+                @test tensors(ψ, 1) isa Tensor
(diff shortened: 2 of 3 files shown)


mofeing:
There is a reason why `tensors` is only defined for `State`s and not `Operator`s, being that it is not as easy for `Operator`s when the input and output sites do not match. This is the case of the Spaced Matrix Product Operator, for example.

I'm closing this PR without merging because I don't think this should be the solution, but thanks for reminding that I have yet to fix this.

@jofrevalles do you mind opening an issue?

If you need a quick fix, you can use the `interlayer` property of `Quantum` TNs along with the `select` function.

jofrevalles:
Okay. Since right now `MPO` and `PEPO` do only support the same amount of inputs/outputs I thought that was a good idea to implement that. Nevertheless, let's discuss that in an issue.

PR #79, opened 2023-08-01
Implement `Infinite` `Boundary` Conditions for `MatrixProduct` and `ProjectedEntangledPair` Types

### LINKED ISSUE #77: Implement `Infinite` boundary conditions for `MatrixProduct` and `ProjectedEntangledPair`
### Summary
`MatrixProduct` and `ProjectedEntangledPair` currently have `Open` and `Periodic` boundary conditions. We could enhance these structs by adding the `Infinite` boundary condition, in which a cell-like structure is repeated infinitely many times.

jofrevalles:
### Summary

In this PR, we resolve #77 by extending our `MatrixProduct` and `ProjectedEntangledPair` types to include support for infinite boundary conditions, materialized in the infinite Matrix Product States (iMPS) and infinite Projected Entangled Pair States (iPEPS). This creates a structure where the tensors cycle periodically, emulating an infinite system by using a finite system unit cell that repeats itself. 

The `tensors`, `Base.show` and `Base.length` functions have been extended to support `TensorNetworks` with `Infinte` boundary conditions. Additional unit tests have been added to ensure the proper functioning of these new boundary conditions for both the `MPS` and `MPO` structures.

Nevertheless, the visualization of infinite boundary types is not yet supported and we should think about how to implement this.

### Example
Let's create a `MatrixProduct{State, Infinite}` which consists of a cell of three tensors that repeat indefinitely:
```julia
julia> _arrays = [rand(1, 1, 2), rand(1, 1, 2), rand(1, 1, 2)]
3-element Vector{Array{Float64, 3}}:
 [0.17136137178613753;;; 0.1659622523369414]
 [0.23387611601182623;;; 0.5187233455849274]
 [0.025368244393775274;;; 0.2701565528371912]

julia> ψ = MatrixProduct{State,Infinite}(_arrays, order = (:l, :r, :o))
TensorNetwork{MatrixProduct{State, Infinite}, NamedTuple{(:plug, :interlayer, :χ), Tuple{Type{<:Plug}, Vector{Bijection{Int64, Symbol}}, Union{Nothing, Int64}}}}(#tensors=∞, #labels=∞)

julia> tensors(ψ, 1) == tensors(ψ, 4) == tensors(ψ, -2)
true
```

### DIFF
--- a/src/Quantum/MP.jl
+++ b/src/Quantum/MP.jl
@@ -46,6 +46,7 @@ _sitealias(::Type{MatrixProduct{P,Open}}, order, n, i) where {P<:Plug} =
         order
     end
 _sitealias(::Type{MatrixProduct{P,Periodic}}, order, n, i) where {P<:Plug} = tuple(order...)
+_sitealias(::Type{MatrixProduct{P,Infinite}}, order, n, i) where {P<:Plug} = tuple(order...)
 
 defaultorder(::Type{MatrixProduct{State}}) = (:l, :r, :o)
 defaultorder(::Type{MatrixProduct{Operator}}) = (:l, :r, :i, :o)
@@ -109,6 +110,17 @@ end
 
 const MPS = MatrixProduct{State}
 const MPO = MatrixProduct{Operator}
+
+tensors(ψ::TensorNetwork{MatrixProduct{P,Infinite}}, args...) where {P<:Plug} =
+    throw(ArgumentError("You need to specify the site for an infinite MatrixProduct$P"))
+
+tensors(ψ::TensorNetwork{MatrixProduct{P,Infinite}}, site::Int, args...) where {P<:Plug} =
+    tensors(plug(ψ), ψ, mod(site, length(ψ)) + 1, args...)
+
+Base.show(io::IO, ψ::TensorNetwork{MatrixProduct{P,Infinite}}) where {P<:Plug} =
+    print(io, "$(typeof(ψ))(#tensors=∞, #labels=∞)")
+
+Base.length(ψ::TensorNetwork{MatrixProduct{P,Infinite}}) where {P<:Plug} = Inf
 
 # NOTE does not use optimal contraction path, but "parallel-optimal" which costs x2 more
 # function contractpath(a::TensorNetwork{<:MatrixProductState}, b::TensorNetwork{<:MatrixProductState})
--- a/src/Quantum/MP.jl
+++ b/src/Quantum/MP.jl
@@ -112,7 +112,7 @@ const MPS = MatrixProduct{State}
 const MPO = MatrixProduct{Operator}
 
 tensors(ψ::TensorNetwork{MatrixProduct{P,Infinite}}, args...) where {P<:Plug} =
-    throw(ArgumentError("You need to specify the site for an infinite MatrixProduct$P"))
+    throw(MethodError("You need to specify the site for an infinite MatrixProduct$P"))
 
 tensors(ψ::TensorNetwork{MatrixProduct{P,Infinite}}, site::Int, args...) where {P<:Plug} =
     tensors(plug(ψ), ψ, mod1(site, length(ψ.tensors)), args...)
--- a/src/Quantum/MP.jl
+++ b/src/Quantum/MP.jl
@@ -115,7 +115,7 @@ tensors(ψ::TensorNetwork{MatrixProduct{P,Infinite}}, args...) where {P<:Plug} =
     throw(ArgumentError("You need to specify the site for an infinite MatrixProduct$P"))
 
 tensors(ψ::TensorNetwork{MatrixProduct{P,Infinite}}, site::Int, args...) where {P<:Plug} =
-    tensors(plug(ψ), ψ, mod(site, length(ψ)) + 1, args...)
+    tensors(plug(ψ), ψ, mod1(site, length(ψ.tensors)), args...)
 
 Base.show(io::IO, ψ::TensorNetwork{MatrixProduct{P,Infinite}}) where {P<:Plug} =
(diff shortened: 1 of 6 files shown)


src/Quantum/MP.jl
```diff
@@ -110,6 +111,17 @@ end
 const MPS = MatrixProduct{State}
 const MPO = MatrixProduct{Operator}
 
+tensors(ψ::TensorNetwork{MatrixProduct{P,Infinite}}, args...) where {P<:Plug} =
+    throw(ArgumentError("You need to specify the site for an infinite MatrixProduct$P"))
+
```
mofeing, line 116: If it doesn't match any method signature, it's better to let Julia throw a `MethodError`.
```suggestion
```

src/Quantum/PEP.jl
```diff
@@ -116,6 +117,17 @@ end
 const PEPS = ProjectedEntangledPair{State}
 const PEPO = ProjectedEntangledPair{Operator}
 
+tensors(ψ::TensorNetwork{ProjectedEntangledPair{P,Infinite}}, args...) where {P<:Plug} =
+    throw(ArgumentError("You need to specify the site for an infinite MatrixProduct$P"))
+
```
mofeing, line 122: Same as in `MP.jl`
```suggestion
```

test/MatrixProductOperator_test.jl
```diff
@@ -76,6 +76,14 @@
...
                 end
             end
+
+            @testset "tensors" begin
+                arrays = [rand(1, 2, 2), rand(1, 1, 2, 2), rand(1, 2, 2)]
+                ψ = MatrixProduct{Operator,Open}(arrays, order = (:l, :r, :i, :o))
+
+                @test tensors(ψ, 1) isa Tensor
+                @test size(tensors(ψ)) |> first == length(ψ) == 3
+            end
```
mofeing, line 86: Again you mixed PRs.
```suggestion
```

src/Quantum/MP.jl
jofrevalles: Sorry, I `cherry-picked` the commits in #78 because I thought was necessary, I will try to remove them now.

src/Quantum/MP.jl
```diff
@@ -110,6 +111,17 @@ end
 const MPS = MatrixProduct{State}
 const MPO = MatrixProduct{Operator}
 
+tensors(ψ::TensorNetwork{MatrixProduct{P,Infinite}}, args...) where {P<:Plug} =
+    throw(MethodError("You need to specify the site for an infinite MatrixProduct$P"))
+
```
mofeing, line 116: I meant don't even put a method jajaja. Sorry if I didn't explain it well.

The Julia Runtime will already throw a `MethodError` if a matching method doesn't exist.

```suggestion
```

src/Quantum/PEP.jl
```diff
@@ -116,6 +117,17 @@ end
 const PEPS = ProjectedEntangledPair{State}
 const PEPO = ProjectedEntangledPair{Operator}
 
+tensors(ψ::TensorNetwork{ProjectedEntangledPair{P,Infinite}}, args...) where {P<:Plug} =
+    throw(MethodError("You need to specify the site for an infinite MatrixProduct$P"))
+
```
mofeing, line 122:
```suggestion
```

src/Quantum/MP.jl
```diff
@@ -110,6 +111,17 @@ end
...
+tensors(ψ::TensorNetwork{MatrixProduct{P,Infinite}}, site::Int, args...) where {P<:Plug} =
+    tensors(plug(ψ), ψ, mod1(site, length(ψ.tensors)), args...)
+
```
mofeing, line 119: This should already be provided by the `tensors` method in `Quantum.jl`. Why reimplement this?
```suggestion
```

src/Quantum/MP.jl
```diff
@@ -110,6 +111,17 @@ end
...
+Base.show(io::IO, ψ::TensorNetwork{MatrixProduct{P,Infinite}}) where {P<:Plug} =
+    print(io, "$(typeof(ψ))(#tensors=∞, #labels=∞)")
+
```
mofeing, line 122: I think this method shouldn't be overridden, but just the `length` method specialized. Maybe the number of labels it's difficult to obtain?
```suggestion
```

src/Quantum/PEP.jl
```diff
@@ -116,6 +117,17 @@ end
...
+tensors(ψ::TensorNetwork{ProjectedEntangledPair{P,Infinite}}, site::Int, args...) where {P<:Plug} =
+    tensors(plug(ψ), ψ, mod1(site, length(ψ.tensors)), args...)
+
+Base.show(io::IO, ψ::TensorNetwork{ProjectedEntangledPair{P,Infinite}}) where {P<:Plug} =
+    print(io, "$(typeof(ψ))(#tensors=∞, #labels=∞)")
+
```
mofeing, line 128: Same as in MP.jl
```suggestion
```

mofeing (changes requested):

src/Quantum/MP.jl
jofrevalles: Well, this is for the specific case of an `Infinite` chain of tensors. For example, if we have two tensors that repeat indefinitely, then `tensors(3)` should return `tensors(1)`. This is different from the `Open` and `Periodic` boundary conditions.

src/Quantum/MP.jl
jofrevalles: Yes, that's it. Since `labels=length(tn.indices)` it was difficult to fix that without overwritting it. However, I think we can change `labels=length(labels(tn))` and this should be fixed.

mofeing (approved):

PR #94, opened 2023-09-18
Add `lu` decomposition for `Tensor`s

### LINKED ISSUE #26: Implement LU factorization for `Tensor`

jofrevalles:
### Summary
This PR adds the new `lu` function for tensors, extending the `LinearAlgebra.lu` function (resolve #26). The new `lu` function returns the LU decomposition of a `Tensor`, where the tensor can be recovered by contracting the permutation tensor `P`, the tensor `L`, and tensor `U`. The tensors `L` and `U` are reshaped versions of the original lower and upper triangular matrices obtained during the decomposition process, respectively.

This implementation is inspired by the LU decomposition in the `scipy` library, as it returns the permutation tensor `P` allowing the original tensor `A` to be recovered with the contraction `A = P * L * U`. This contrasts with `LinearAlgebra`, where the permutation vector `p` is returned, and the original matrix can be recovered with `P' * A = L * U` (where `P'` is the permutation matrix built from `p`).

Please let me know if there are any concerns or issues with extending the `LinearAlgebra` library in this manner.

We have also added tests for this new function.

### Example
A usage example of the lu function:
```julia
julia> using Tenet; using LinearAlgebra; using Test

julia> tensor = Tensor(rand(4, 4, 4), (:i, :j, :k))
4×4×4 Tensor{Float64, 3, Array{Float64, 3}}:
...

julia> P, L, U = lu(tensor, left_inds = labels(tensor)[1:2])
...

julia> @test contract(contract(P, L), U) ≈ tensor
Test Passed
```

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -14,6 +14,7 @@ LinearAlgebra = "37e2e46d-f89d-539d-b4ee-838fcccc9c8e"
 Muscle = "21fe5c4b-a943-414d-bf3e-516f24900631"
 OMEinsum = "ebe7aa44-baf0-506c-a96f-8464559b3922"
 Random = "9a3f8284-a2c9-5f02-9a11-845980a1fd5c"
+SparseArrays = "2f01184e-e22b-5df5-ae63-d93ebab69eaf"
 UUIDs = "cf7118a7-6976-5b1a-9a39-7adc72f591a4"
 ValSplit = "0625e100-946b-11ec-09cd-6328dd093154"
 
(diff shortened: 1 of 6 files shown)


mofeing:
@jofrevalles I've updated the code to the recent changes. Also, I've refactored the SVD and QR factorizations. Would you mind giving a quick view?

jofrevalles:
> @jofrevalles I've updated the code to the recent changes. Also, I've refactored the SVD and QR factorizations. Would you mind giving a quick review?

I have checked it and everything looks good to me!

mofeing (approved):

PR #97, opened 2023-09-19
Refactor `Ansatz` to traits

mofeing:
WIP


src/TensorNetwork.jl
```diff
@@ -5,88 +5,92 @@ using OMEinsum
...
-metadata(T::Type{<:Arbitrary}) = metadata(supertype(T))
+Return the `Domain` of a [`TensorNetwork`](@ref).
+"""
+domain(::TensorNetwork{D}) where {D} = D
+domain(::Type{<:TensorNetwork{D}}) where {D} = D
 
-Base.summary(io::IO, x::TensorNetwork) = print(io, "$(length(x))-tensors $(typeof(x))")
-Base.show(io::IO, tn::TensorNetwork) = print(io, "$(typeof(tn))(#tensors=$(length(tn)), #inds=$(length(tn.indices)))")
+# TODO do sth to skip `checkdomain`? like `@inbounds`
+checkdomain(tn::TensorNetwork{Domain}) = nothing
```
jofrevalles, line 74: We have to include the `NamedTuple` to be able to call `checkdomain` with a `TensorNetwork`:
```suggestion
checkdomain(tn::TensorNetwork{D, F}) where {D<:Domain, F<:NamedTuple} = nothing
```

jofrevalles (changes requested):

src/TensorNetwork.jl
mofeing: Nope, this must not work for subtypes of `Domain`. Only for `TensorNetwork{Domain}` itself. And `TensorNetwork{Domain,NamedTuple{(),Tuple{}}} <: TensorNetwork{Domain} <: TensorNetwork`, so no need to specify `F`.

Check out the `inheritcheckdomain` function. It recursively calls `checkdomain` on supertypes until it arrives `Domain` supertype.

PR #103, opened 2023-10-06
Refactor `TensorNetwork` to class-based OOP organization

### LINKED ISSUE #96: Refactor TN `Ansatz`es to traits
Using type-parameters for encoding TN ansatz is not going to scale well.

1. `Composite` incurs in a lot of compilation.
2. Runtime dispatch is more costly than conditionals.

## Proposal

The idea is to replace `MatrixProduct`, `ProjectedEntangledPair`, ... for Holy traits.

```julia
abstract type MatrixProduct{P,B} <: Quantum where {P<:Plug,B<:Boundary} end
```

would be rewritten as

```julia
abstract type Ansatz end
struct MatrixProduct{P<:Plug,B<:Boundary} <: Ansatz end
```

Also, the current `Ansatz` type would be renamed, because it would no longer express what we are meaning about it (i.e. the form of the graph). I suggest renaming it to `Domain` because that's what its fields are going to vary on.

## Links

https://invenia.github.io/blog/2019/11/06/julialang-features-part-2/
https://www.ahsmart.com/pub/holy-traits-design-patterns-and-best-practice-book/

### LINKED ISSUE #102: Rename `append!` to `merge!`
`Base.append!` is thought for ordered collections like `Vector`, while `Base.merge!` is thought for key-indexed collections like `Dict`. Our `TensorNetwork` type resembles more a `Dict` since `Tensor`s are not ordered and are key-indexed, so semantically makes more sense to use `merge!`.

mofeing:
This PR starts the process of rewriting current `TensorNetwork` type hierarchy on more OOP-like classes. It doesn't do all the things we would like from it yet, but we need some other changes and this PR merges the minimum changes in order to proceed with them.

The main change is that we get rid of the "field-inheritance hack using `NamedTuple`", which although interesting, it doesn't scale well. It also redefines what an `Ansatz` is and turns `Boundary` and `Plug` into traits.

# Things yet to do after this PR

- [ ] Refactor `TensorNetwork` internals to encode the dual graph of the Tensor Network #104 
- [ ] Implement "multilayer-graph"-like functionality in `QuantumTensorNetwork` #105 
- [ ] Design `Ansatz` subtypes as special cases of graph layers #106 
  - Graph layers should encode the entanglement connectivity pattern (i.e. virtual indices)
  - The `Ansatz` **IS** the graph layer (e.g. a MPS and a MPO have the same "ansatz" and graph layer)
- [ ] Consider deprecating `Infinite` plug trait #107 
  - Infinite graphs are difficult to manage
  - Many "infinite" algorithms can actually be implemented by using "open" tensor networks
- [ ] Discuss composition over field inheritance for the `TensorNetwork` hierarchy
  - Currently we are doing _field inheritance_ with `Classes.@class`
- [ ] Consider a `Classes.jl` fork or build on top of other packages #108

### DIFF
--- a/docs/src/transformations.md
+++ b/docs/src/transformations.md
@@ -81,7 +81,7 @@ A = Tensor(data, (:i, :j, :k, :l)) #hide
 B = Tensor(rand(2, 2), (:i, :m)) #hide
 C = Tensor(rand(2, 2), (:j, :n)) #hide
 
-tn = TensorNetwork(Tensor[A, B, C]) #hide
+tn = TensorNetwork([A, B, C]) #hide
 reduced = transform(tn, Tenet.DiagonalReduction) #hide
 
 smooth_annotation!( #hide
@@ -139,7 +139,7 @@ B = Tensor(rand(2, 2), (:i, :m)) #hide
 C = Tensor(rand(2, 2, 2), (:m, :n, :o)) #hide
 E = Tensor(rand(2, 2, 2, 2), (:o, :p, :q, :j)) #hide
 
-tn = TensorNetwork(Tensor[A, B, C, E]) #hide
+tn = TensorNetwork([A, B, C, E]) #hide
 reduced = transform(tn, Tenet.RankSimplification) #hide
 
 smooth_annotation!( #hide
@@ -193,7 +193,7 @@ A = Tensor(data, (:i, :j, :k)) #hide
 B = Tensor(rand(3, 3), (:j, :l)) #hide
 C = Tensor(rand(3, 3), (:l, :m)) #hide
 
-tn = TensorNetwork(Tensor[A, B, C]) #hide
+tn = TensorNetwork([A, B, C]) #hide
 reduced = transform(tn, Tenet.ColumnReduction) #hide
 
 smooth_annotation!( #hide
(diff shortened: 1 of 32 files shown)


src/TensorNetwork.jl
```diff
@@ -3,109 +3,57 @@ using Random
...
-# TODO maybe rename it as `convert` method?
-TensorNetwork{A}(tn::TensorNetwork{B}; metadata...) where {A,B} =
-    TensorNetwork{A}(tensors(tn); merge(tn.metadata, metadata)...)
-
-# TODO do sth to skip checkansatz? like @inbounds
-function checkansatz(tn::TensorNetwork{A}) where {A<:Ansatz}
-    for T in superansatzes(A)
-        checkmeta(T, tn) || throw(ErrorException("\"$T\" metadata is not valid"))
+TensorNetwork() = TensorNetwork(Tensor[])
+function TensorNetwork(tensors)
```
jofrevalles, line 20:
```suggestion
TensorNetwork(tensors::Vector{<:Tensor}) = TensorNetwork(Tensor[tensors...])

function TensorNetwork(tensors::Vector{Tensor})
```

With this change, we will be able to call `TensorNetwork` with a `Vector` of `Tensor`s  without having to explicitly do `Tensor[tensor1, tensor2, ...]`.

jofrevalles (changes requested):

src/TensorNetwork.jl
mofeing: I think it would be better if we just do `convert(Vector{Tensor}, tensors)` inside `TensorNetwork`'s inner constructor.

docs/src/transformations.md
```diff
@@ -81,7 +81,7 @@ A = Tensor(data, (:i, :j, :k, :l)) #hide
 B = Tensor(rand(2, 2), (:i, :m)) #hide
 C = Tensor(rand(2, 2), (:j, :n)) #hide
 
-tn = TensorNetwork([A, B, C]) #hide
+tn = TensorNetwork(Tensor[A, B, C]) #hide
```
jofrevalles, line 84:
```suggestion
tn = TensorNetwork([A, B, C]) #hide
```
We can now maybe remove all these explicit `Tensor` in the `Vector` arguments, so the user does not see any change in the interface. What do you think, @mofeing ?

jofrevalles (changes requested):

jofrevalles (approved):

PR #113, opened 2023-10-26
Enhance `qr` function with `mode` argument

jofrevalles:
### Summary
Currently, the `qr` decomposition on a `Tensor` t defaults to the "full" mode, which given a matrix `(M, N)` with `M>N` we get a `Q` matrix with size `(M, M)` and `R` with size `(N, N)`. This is counter-intuitive since these two tensors share a virtual index but they do not have the same dimensions. 

In this PR we fix this by adding a `mode` argument to the `qr` function (which by default is set to `:reduced`). With `mode=:reduced`, it will ensure that the dimensions of the virtual index match and with `mode=:full` the `qr` functions returns the full matrices.

### Example of usage
This enhancement facilitates more intuitive operations as illustrated below:
```julia
 julia> using Tenet

 julia> A = Tensor(rand(32, 8), (:i, :j))
32×8 Tensor{Float64, 2, Matrix{Float64}}: ...

julia> Q, R = qr(A, :reduced; left_inds=(:i,))
...

julia> size(Q), size(R)
((32, 8), (8, 8))

julia> Q, R = qr(A, :full; left_inds=(:i,))
...

julia> size(Q), size(R)
((32, 32), (8, 8))
```


jofrevalles:
@mofeing I don't know why when trying to merge in `develop` I pick up all these commits

mofeing:
Most probably because you started your branch from `master` and not from `develop`.

jofrevalles:
> Most probably because you started your branch from `master` and not from `develop`.

Let me re-do the PR.

jofrevalles:
Re-opened correctly in #114

PR #114, opened 2023-10-26
Enhance `qr` function with `mode` argument

jofrevalles:
### Summary
Currently, the `qr` decomposition on a `Tensor` t defaults to the "full" mode, which given a matrix `(M, N)` with `M>N` we get a `Q` matrix with size `(M, M)` and `R` with size `(N, N)`. This is counter-intuitive since these two tensors share a virtual index but they do not have the same dimensions. 

In this PR we fix this by adding a `mode` argument to the `qr` function (which by default is set to `:reduced`). With `mode=:reduced`, it will ensure that the dimensions of the virtual index match and with `mode=:full` the `qr` functions returns the full matrices.

### Example of usage
This enhancement facilitates more intuitive operations as illustrated below:
```julia
 julia> using Tenet

 julia> A = Tensor(rand(32, 8), (:i, :j))
32×8 Tensor{Float64, 2, Matrix{Float64}}: ...

julia> Q, R = qr(A, :reduced; left_inds=(:i,))
...

julia> size(Q), size(R)
((32, 8), (8, 8))

julia> Q, R = qr(A, :full; left_inds=(:i,))
...

julia> size(Q), size(R)
((32, 32), (8, 8))
```

### DIFF
--- a/src/Numerics.jl
+++ b/src/Numerics.jl
@@ -129,7 +129,7 @@ Perform QR factorization on a tensor.
     - `right_inds`: right indices to be used in the QR factorization. Defaults to all indices of `t` except `left_inds`.
     - `virtualind`: name of the virtual bond. Defaults to a random `Symbol`.
 """
-function LinearAlgebra.qr(t::Tensor, mode::Symbol = :reduced; left_inds = (), right_inds = (), virtualind::Symbol = Symbol(uuid4()), kwargs...)    isdisjoint(left_inds, right_inds) ||
+function LinearAlgebra.qr(t::Tensor; left_inds = (), right_inds = (), virtualind::Symbol = Symbol(uuid4()), kwargs...)    isdisjoint(left_inds, right_inds) ||
         throw(ArgumentError("left ($left_inds) and right $(right_inds) indices must be disjoint"))
 
     left_inds, right_inds =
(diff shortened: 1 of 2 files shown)


mofeing:
I see the problem but I don't think this is the solution. In the documentation of `LinearAlgebra.qr` function you can find the following paragraph:

> ...
> Multiplication with respect to either full/square or non-full/square `Q` is allowed, i.e. both `F.Q*F.R` and `F.Q*A` are supported. A `Q` matrix can be converted into a regular matrix with `Matrix`. **This operation returns the "thin" `Q` factor**, i.e., if `A` is `m`×`n` with `m>=n`, then **`Matrix(F.Q)` yields an m×n matrix** with orthonormal columns. To retrieve the "full" `Q` factor, an m×m orthogonal matrix, use `F.Q*I`. If `m<=n`, then `Matrix(F.Q)` yields an m×m orthogonal matrix.
> ...

I would default to retrieving the thin `Q` matrix without a `mode` argument.

jofrevalles:
> I would default to retrieving the thin `Q` matrix without a `mode` argument.

Okay, this also makes sense. In this way we always get `size(Q, virtualind)=size(R, virtualind)` which follows our implementation.

mofeing:
Is the PR ready?

mofeing (approved):

PR #120, opened 2023-11-06
Refactor `TensorNetwork` internals to incidence matrix representation

mofeing:
This PR refactors the `TensorNetwork` internals to be more like a incidence matrix.


test/Transformations_test.jl
```diff
@@ -188,57 +188,36 @@
...
-                @test isempty(Tenet.find_zero_columns(parent(tensor)))
-                @test size(tensor, :j) == 2
-            end
-
-            @test length(tn.indices) == length(reduced.indices)
-
-            # Test that the resulting contraction is the same as the original
-            @test contract(reduced) ≈ view(contract(tn), :j => 1:2:3)
+            @test size(reduced, :j) == 2
+            @test contract(reduced) ≈ contract(tn)
```
jofrevalles, line 220: Okay, be careful here since if the `:j` index is not contracted, this line would not work and we would need something like `contract(reduced) ≈ view(contract(tn), :j => 1:2:3)`.

src/Transformations.jl
```diff
@@ -215,56 +213,14 @@ end
 function transform!(tn::AbstractTensorNetwork, config::ColumnReduction)
     skip_inds = isempty(config.skip) ? inds(tn, set = :open) : config.skip
```
jofrevalles, line 214:
```suggestion
    skip_inds = config.skip
```
Couldn't we just put this here in the `ColumnReduction` struct definition? The user may want to actually have the open indices reduced.

jofrevalles (changes requested):

src/Transformations.jl
mofeing: Nope, because at the struct definition you don't have the tensor network yet.

test/Transformations_test.jl
mofeing: I don't see why. We designed the case and we are setting to contract because it appears more than once in `tn`.

PR #124, opened 2023-11-27
Fix excessive synchronization in `contract` on `Dagger.DArray`

### LINKED ISSUE #123: Specialize `contract` method for `Dagger.DArray` tensors
When the tensors to be contracted are matrices, the generated tasks are executed in parallel, and the total number of tasks makes sense (first image). The problem arises when the tensors to be contracted are of higher order, as EinCode starts to perform permutations in the dimensions and reshapes to carry out the contraction, seemingly generating an endless number of tasks (second image, and it hasn't even reached the end of the contraction). In the case where the tensors were matrices, these operations are not necessary, and the contraction operation is merely the matrix product, yielding the expected result.

two_matrices_contraction_darrays

3-order_tensor_contraction_darrays

mofeing:
Closes #123


jofrevalles (commented):
Can we add some tests? :)

Todorbsc:
> Can we add some tests? :)

Yeah, we should. Currently, we are trying to fix some things in order to make the code work.

mofeing:
Updated the code to the last `Tenet` and `Dagger` changes. Also it now calls `Tenet.contract` instead of `OMEinsum.EinCode`.

@jofrevalles I also added some tests.

test/integration/Dagger_test.jl
```diff
@@ -0,0 +1,60 @@
...
+                contracted_tensor = contract(tensor1, tensor2)
+                contracted_block_tensor = contract(block_tensor1, block_tensor2)
+
+                @test parent(contracted_block_tensor) isa DArray
+                @test inds(contracted_block_tensor) == [:i, :k]
+                @test all(==((2, 2)) ∘ size, Dagger.domainchunks(parent(contracted_block_tensor)))
+                @test collect(parent(contracted_block_tensor)) ≈ parent(contracted_tensor)
+            end
+
+            #     @testset "block-unblock" begin
```
jofrevalles, line 42: Why is this commented?

mofeing: Because calling `contract` on a `Dagger.DArray`-backed `Tensor` and a non-`DArray` `Tensor` is not yet supported. But I wanted to keep it.

Actually, I think it is a very edgy case: IMO it is reasonable to ask users to convert first all tensors to `Tensor{DArray}`s before calling `contract`.

jofrevalles (approved):

Todorbsc (approved):

PR #137, opened 2024-03-20
Fix sum and subtraction for `Tensor` with same `inds` but different permutation

### LINKED ISSUE #136: Basic arithmetic operation fails on `Tensor`s with matching `inds` but different permutation
When attempting to perform a sum or subtraction between two tensors that have matching `inds` but in a different permutation, the operation fails with a `DimensionMismatch` error. 
I instead expected that these operations were possible.

Example:
```julia
julia> using Tenet

julia> A = Tensor(rand(2, 4, 2), (:i, :j, :k))
2×4×2 Tensor{Float64, 3, Array{Float64, 3}}:
[:, :, 1] =
 0.864098  0.504478  0.764864  0.576408
 0.17088   0.644568  0.889588  0.883436

[:, :, 2] =
 0.0782172  0.893183    0.507304  0.758659
 0.836397   0.00498142  0.103803  0.176029

julia> B = Tensor(rand(4, 2, 2), (:j, :i, :k))
4×2×2 Tensor{Float64, 3, Array{Float64, 3}}:
[:, :, 1] =
 0.932157  0.505743
 0.865242  0.1574
 0.290655  0.269668
 0.267595  0.609859

[:, :, 2] =
 0.675176  0.725486
 0.737603  0.377162
 0.254187  0.925577
 0.980923  0.207142

julia> A - B
ERROR: DimensionMismatch: dimensions must match: a has dims (Base.OneTo(2), Base.OneTo(4), Base.OneTo(2)), b has dims (Base.OneTo(4), Base.OneTo(2), Base.OneTo(2)), mismatch at 1
Stacktrace:
 [1] promote_shape
   @ ./indices.jl:178 [inlined]
 [2] promote_shape(a::Tensor{Float64, 3, Array{Float64, 3}}, b::Tensor{Float64, 3, Array{Float64, 3}})
   @ Base ./indices.jl:169
 [3] -(A::Tensor{Float64, 3, Array{Float64, 3}}, B::Tensor{Float64, 3, Array{Float64, 3}})
   @ Base ./arraymath.jl:7
 [4] top-level scope
   @ REPL[5]:1
```

jofrevalles:
This PR resolves #136. Additionally, we add some small tests to cover for the changes.


test/Numerics_test.jl
```diff
@@ -1,6 +1,19 @@
...
+        A = Tensor(rand(2, 3, 2), (:i, :j, :k))
+        B = Tensor(rand(3, 2, 2), (:j, :i, :k))
+
+        C = A + B
+        @test issetequal(inds(C), (:i, :j, :k))
+        @test issetequal(size(C), (2, 3, 2))
+
+        D = A - B
+        @test issetequal(inds(D), (:i, :j, :k))
+        @test issetequal(size(D), (2, 3, 2))
```
mofeing, line 14: You should also test the result.

src/Numerics.jl
```diff
@@ -33,6 +33,18 @@ end
...
+function Base.:(+)(a::Tensor, b::Tensor)
+    issetequal(inds(a), inds(b)) || return false
+    perm = __find_index_permutation(inds(a), inds(b))
+    return Tensor(parent(a) + permutedims(parent(b), perm), inds(a))
+end
+
+function Base.:(-)(a::Tensor, b::Tensor)
+    issetequal(inds(a), inds(b)) || return false
+    perm = __find_index_permutation(inds(a), inds(b))
+    return Tensor(parent(a) + permutedims(parent(b), perm), inds(a))
```
mofeing, line 45: Maybe better to use `PermuteDimsArray` to avoid excessive copying (it creates a view instead of copying `b`).
```suggestion
    return Tensor(parent(a) + PermuteDimsArray(parent(b), perm), inds(a))
end

function Base.:(-)(a::Tensor, b::Tensor)
    issetequal(inds(a), inds(b)) || return false
    perm = __find_index_permutation(inds(a), inds(b))
    return Tensor(parent(a) + PermuteDimsArray(parent(b), perm), inds(a))
```

mofeing (approved):

src/Numerics.jl
```diff
@@ -33,6 +33,18 @@ end
 # NOTE use `String[...]` code instead of `map` or broadcasting to set eltype in empty cases
 __omeinsum_sym2str(x) = String[string(i) for i in x]
 
+function Base.:(+)(a::Tensor, b::Tensor)
+    issetequal(inds(a), inds(b)) || return false
```
mofeing, line 37: This should error.

src/Numerics.jl
```diff
@@ -33,6 +33,18 @@ end
...
+    perm = __find_index_permutation(inds(a), inds(b))
+    return Tensor(parent(a) + PermutedDimsArray(parent(b), perm), inds(a))
+end
+
+function Base.:(-)(a::Tensor, b::Tensor)
+    issetequal(inds(a), inds(b)) || return false
```
mofeing, line 43: And this too.

mofeing (changes requested):

src/Numerics.jl
```diff
@@ -33,6 +33,18 @@ end
...
 __omeinsum_sym2str(x) = String[string(i) for i in x]
 
+function Base.:(+)(a::Tensor, b::Tensor)
+    issetequal(inds(a), inds(b)) || raise(ArgumentError("indices must be equal"))
+    perm = __find_index_permutation(inds(a), inds(b))
+    return Tensor(parent(a) + PermutedDimsArray(parent(b), perm), inds(a))
+end
+
+function Base.:(-)(a::Tensor, b::Tensor)
+    issetequal(inds(a), inds(b)) || raise(ArgumentError("indices must be equal"))
```
mofeing, line 43: There is no `raise` in Julia. I think you confused it with `throw`.
```suggestion
    issetequal(inds(a), inds(b)) || throw(ArgumentError("indices must be equal"))
    perm = __find_index_permutation(inds(a), inds(b))
    return Tensor(parent(a) + PermutedDimsArray(parent(b), perm), inds(a))
end

function Base.:(-)(a::Tensor, b::Tensor)
    issetequal(inds(a), inds(b)) || throw(ArgumentError("indices must be equal"))
```

mofeing (changes requested):

src/Numerics.jl
jofrevalles: oops! Yes!

mofeing (approved):

jofrevalles:
Can we merge this, @mofeing ?

mofeing:
Yep, I approved it

PR #35, opened 2023-03-09
Add `MatrixProductOperator` type

emapuljak:
Creation of MatrixProductOperator and related tests.

PR #169, opened 2024-07-09
Extend `evolve` of an `MPS` with an `MPO`

jofrevalles:
### Summary
This PR was left open in the legacy code in `Qrochet` ([#35](https://github.com/bsc-quantic/Qrochet.jl/pull/35)). This PR extends the `evolve` function, which evolves a Matrix Product State (`MPS`) using a Matrix Product Operator (`MPO`). The idea from @starsfordummies is that we `contract` the tensors of each corresponding `site`, and we then create the new `λ` with a kron product with the old `λ` (which come from the `MPS`) and the identity (comes from the `MPO`). Right now, this function only works for mps in the canonical form.

### Example
Here we show how can we use this function:

```julia
julia> using Tenet

julia> ψ = rand(Chain, Open, State; n=8, χ=10); canonize!(ψ)
MPS (inputs=0, outputs=8)

julia> mpo = rand(Chain, Open, Operator; n=8, χ=10)
MPO (inputs=8, outputs=8)

julia> Tenet.@reindex! inputs(mpo) => outputs(ψ)
MPS (inputs=0, outputs=8)

julia> ψ = Tenet.evolve(ψ, mpo)
MPS (inputs=0, outputs=8)

julia> size.(tensors(ψ))
15-element Vector{Tuple{Int64, Vararg{Int64}}}:
 (8,)
 (40,)
 (80,)
 (100,)
 (80,)
 (40,)
 (8,)
 (2, 8)
 (2, 8, 40)
 (2, 40, 80)
 (2, 80, 100)
 (2, 100, 80)
 (2, 80, 40)
 (2, 40, 8)
 (2, 8)
```

### DIFF
--- a/src/Ansatz/Chain.jl
+++ b/src/Ansatz/Chain.jl
@@ -712,6 +712,64 @@ function unpack_2sitewf!(ψ::Chain, bond, left_inds, right_inds, virtualind)
     return ψ
 end
 
+Tenet.__check_index_sizes(qtn::Chain) = Tenet.__check_index_sizes(TensorNetwork(qtn))
+
+evolve(ψ::Chain, mpo::Chain) = evolve!(copy(ψ), mpo)
+
+"""
+    evolve!(ψ::Chain, mpo::Chain)
+
+Applies a Matrix Product Operator (MPO) `mpo` to the [`Chain`](@ref).
+"""
+function evolve!(ψ::Chain, mpo::Chain)
+    L = nsites(ψ)
+
+    Tenet.@unsafe_region ψ begin
+        for i in 1:L
+            contractedind = inds(ψ; at = Site(i))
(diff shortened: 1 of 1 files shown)


mofeing:
Yep, like I said in the previous PR, those virtual bond dimensions are a big problem.

jofrevalles:
> Yep, like I said in the previous PR, those virtual bond dimensions are a big problem.

I think that we just have to truncate.

mofeing:
> > Yep, like I said in the previous PR, those virtual bond dimensions are a big problem.
> 
> I think that we just have to truncate.

Great, that should solve it.

jofrevalles:
I am closing this, since PR #264 already introduced this feature. @starsfordummies Nevertheless, this code has that idea of extending the canonical form of an `MPS` with $\lambda \otimes I$ that you proposed. It seems interesting, we need to check it.

PR #170, opened 2024-07-09
Replace the global index counter generator for a local data structure

mofeing:
There have been reports that when generating too many indices automatically (using the atomic counter) in the same Julia session, it breaks.

This usually happens due to repeatedly generating `TensorNetworks`.

This PR fixes it by introducing `IndexCounter`: a small wrapper over the currently global behaviour.
If merged, all `TensorNetwork`s would start from the same index.


jofrevalles (approved):
👍

PR #175, opened 2024-07-31
Replace Gram-Schmidt method by QR for random Chain initialization

jofrevalles:
In this PR, we replace the former Gram-Schmidt method for random chain initialization by a `QR` method, which applies the `qr` decomposition to some random tensors in order to get a randomly normalized right-canonized `MPS`.

Additionally, we benchmarked both methods and we conclude that the new method outscales the former in terms of running time, memory required and allocations produced.

### Comparison

· `Gram-Schmidt` method:
```julia
julia> using Tenet; using BenchmarkTools

julia> @btime mps = rand(Chain, Open, State; n=50, χ=300)
  2.002 s (53812 allocations: 122.75 MiB)
MPS (inputs=0, outputs=50)


julia> @btime mpo = rand(Chain, Open, Operator; n=50, χ=200)
  1.537 s (63338 allocations: 120.70 MiB)
MPO (inputs=50, outputs=50)
```

· `QR` method:
```julia
julia> using Tenet; using BenchmarkTools

julia> @btime mps = rand(Chain, Open, State; n=50, χ=300)
  423.495 ms (10480 allocations: 101.18 MiB)
MPS (inputs=0, outputs=50)

julia> @btime mpo = rand(Chain, Open, Operator; n=50, χ=200)
  573.715 ms (28459 allocations: 109.77 MiB)
MPO (inputs=50, outputs=50)
```

### DIFF
--- a/src/Ansatz/Chain.jl
+++ b/src/Ansatz/Chain.jl
@@ -229,32 +229,24 @@ function Base.rand(rng::Random.AbstractRNG, sampler::ChainSampler, ::Type{Open},
     p = get(sampler.parameters, :p, 2)
     T = get(sampler.parameters, :eltype, Float64)
 
-    arrays = Vector{AbstractArray{T,N} where {N}}()
+    arrays::Vector{AbstractArray{T,N} where {N}} = map(1:n) do i
+        χl, χr = let after_mid = i > n ÷ 2, i = (n + 1 - abs(2i - n - 1)) ÷ 2
+            χl = min(χ, p^(i - 1))
+            χr = min(χ, p^i)
 
-    # Left boundary tensor
-    F = lq!(rand(rng, T, 1, p * min(χ, p)))
-    push!(arrays, reshape(Matrix(F.Q), p, min(χ, p)))
-
-    # Bulk tensors
-    for i in 2:(n - 1)
-        χl = min(χ, p^(i - 1))
-        χr = min(χ, p^i)
-
-        if isodd(n) && i == n ÷ 2 + 1
-            χr = χl
-        elseif i > n ÷ 2
-            j = (n + 1 - abs(2i - n - 1)) ÷ 2
-            χl = min(χ, p^j)
-            χr = min(χ, p^(j - 1))
+            # swap bond dims after mid and handle midpoint for odd-length MPS
+            (isodd(n) && i == n ÷ 2 + 1) ? (χl, χl) : (after_mid ? (χr, χl) : (χl, χr))
         end
 
+        # orthogonalize by Gram-Schmidt algorithm
         F = lq!(rand(rng, T, χl, p * χr))
-        push!(arrays, reshape(Matrix(F.Q), χl, p, χr))
+
+        reshape(Matrix(F.Q), χl, p, χr)
     end
 
-    # Right boundary tensor
-    F = lq!(rand(rng, T, min(χ, p), p))
-    push!(arrays, Matrix(F.Q))
+    # reshape boundary sites
+    arrays[1] = reshape(arrays[1], p, p)
+    arrays[n] = reshape(arrays[n], p, p)
 
     return Chain(State(), Open(), arrays; order=(:l, :o, :r))
 end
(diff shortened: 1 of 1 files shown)


src/Ansatz/Chain.jl
```diff
@@ -223,42 +222,43 @@ function Base.rand(rng::AbstractRNG, ::Type{A}, ::Type{B}, ::Type{S}; kwargs...)
...
-        χl, χr = let after_mid = i > n ÷ 2, i = (n + 1 - abs(2i - n - 1)) ÷ 2
-            χl = min(χ, p^(i - 1))
-            χr = min(χ, p^i)
+    arrays = Vector{AbstractArray{T,N} where {N}}()
 
-            # swap bond dims after mid and handle midpoint for odd-length MPS
-            (isodd(n) && i == n ÷ 2 + 1) ? (χl, χl) : (after_mid ? (χr, χl) : (χl, χr))
-        end
+    # Left boundary tensor
+    F = lq!(rand(rng, T, 1, p * min(χ, p)))
```
mofeing, line 235: Why `lq!` and not `qr!`?

mofeing (commented):
Can't you reuse the `map(1:n) do i` block but just changing the line `A = gramschmidt!(rand(rng, T, χl, χr * p))` for `A, _ = qr!(rand(rng, T, χl, χr * p))`.

Is there another reason for refactoring to a `for` loop?

jofrevalles:
> Can't you reuse the `map(1:n) do i` block but just changing the line `A = gramschmidt!(rand(rng, T, χl, χr * p))` for `A, _ = qr!(rand(rng, T, χl, χr * p))`.
> 
> Is there another reason for refactoring to a `for` loop?

We can, but then we would have to put conditions for the first and last tensor, which seems more ugly.

src/Ansatz/Chain.jl
jofrevalles: `lq!` is useful here since we want the `Chain` to be in the right-canonical form. We could do it with `qr!` too but it would involve some permutations and reshapes, which is much elegantly solved with `lq!`.

jofrevalles: If we changed `lq!` for `qr!` and we reshaped the physical leg with the left bond, we would get a left-canonical `Chain`.

src/Ansatz/Chain.jl
```diff
@@ -238,27 +238,19 @@ function Base.rand(rng::Random.AbstractRNG, sampler::ChainSampler, ::Type{Open},
             (isodd(n) && i == n ÷ 2 + 1) ? (χl, χl) : (after_mid ? (χr, χl) : (χl, χr))
         end
 
-        # fix for first site
-        i == 1 && ((χl, χr) = (χr, 1))
-
         # orthogonalize by Gram-Schmidt algorithm
```
mofeing, line 241:
```suggestion
        # orthogonalize by QR factorization
```

src/Ansatz/Chain.jl
```diff
@@ -277,26 +269,16 @@ function Base.rand(rng::Random.AbstractRNG, sampler::ChainSampler, ::Type{Open},
...
 
-        shape = if i == 1
-            (χr, ip, op)
-        elseif i == n
-            (χl, ip, op)
-        else
-            (χl, χr, ip, op)
-        end
-
         # orthogonalize by Gram-Schmidt algorithm
```
mofeing, line 272:
```suggestion
        # orthogonalize by QR factorization
```

src/Ansatz/Chain.jl
```diff
@@ -277,26 +269,16 @@ function Base.rand(rng::Random.AbstractRNG, sampler::ChainSampler, ::Type{Open},
...
 
-    # normalize
-    ζ = min(χ, ip * op)
-    arrays[1] ./= sqrt(ζ)
+    # reshape boundary sites
+    arrays[1] = reshape(arrays[1], p, p, min(χ, ip * op))
+    arrays[n] = reshape(arrays[n], min(χ, ip * op), p, p)
 
-    return Chain(Operator(), Open(), arrays)
+    return Chain(Operator(), Open(), arrays; order=(:l, :i, :o, :r))
```
mofeing, line 281:
```suggestion
    # TODO order might not be the best for performance
    return Chain(Operator(), Open(), arrays; order=(:l, :i, :o, :r))
```

mofeing (changes requested):

PR #185, opened 2024-08-14
Sort output of `sites`

starsfordummies:
just sort() without ! should work here, right?

### DIFF
--- a/src/Quantum.jl
+++ b/src/Quantum.jl
@@ -110,7 +110,7 @@ Returns the sites of a [`Quantum`](@ref) Tensor Network.
 """
 function sites(tn::Quantum; kwargs...)
     if isempty(kwargs)
-        collect(keys(tn.sites))
+      sort(collect(keys(tn.sites)))
     elseif keys(kwargs) === (:at,)
         findfirst(i -> i === kwargs[:at], tn.sites)
     else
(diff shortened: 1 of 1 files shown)


src/Quantum.jl
```diff
@@ -110,7 +110,7 @@ Returns the sites of a [`Quantum`](@ref) Tensor Network.
 """
 function sites(tn::Quantum; kwargs...)
     if isempty(kwargs)
-        collect(keys(tn.sites))
+        sort(collect(keys(tn.sites)))
```
mofeing, line 113: It's better if you use `sort!(collect(...))` to not allocate twice or `sort(keys(...))` to allocate just once when sorting.

```suggestion
        sort(keys(tn.sites))
```

mofeing (changes requested):

src/Quantum.jl
starsfordummies: maybe it's better to just use sort!, if I use your suggestion 
```
> sort(keys(Quantum(psi).sites))
ERROR: MethodError: no method matching sort(::Base.KeySet{Site, Dict{Site, Symbol}})
The function `sort` exists, but no method is defined for this combination of argument types.
```

mofeing: ok, then use `sort!(collect(...))`

mofeing (approved):

PR #186, opened 2024-08-14
Implement `inds` method for physical and virtual indices

mofeing:
With this PR, `inds(tn; set=:physical)` would return the physical indices annotated in the `sites` dict of `Quantum`, and `inds(tn; set=:virtual)` would return the virtual indices (i.e. the set difference between all the indices and the physical indices).


starsfordummies (approved):
thank youu

PR #192, opened 2024-08-20
Lazy builders

starsfordummies:
since we end up writing those a million times in testing, these shorthands might be useful
for me we could also export them


mofeing:
I like the idea of having these kind of constructors, but not this way nor with these names. We can write another proposal when I'm back.

starsfordummies:
ok, note that at the moment these are mostly meant to make life easier on us poor testers, so the more straightforward the better

starsfordummies:
With latest changes and all this is outdated, besides now it's relatively easy to build an MPS with rand()

mofeing:
We can still add a constructor for Pauli strings like `Product("001+")` if you want

PR #199, opened 2024-09-07
Introduce `groupinds!` function to group parallel inds

mofeing:
CC @starsfordummies

### DIFF
--- a/src/TensorNetwork.jl
+++ b/src/TensorNetwork.jl
@@ -644,6 +644,33 @@ This method is equivalent to `tensor ∈ tensors(tn)` code, but it's faster on l
 Base.in(tensor::Tensor, tn::TensorNetwork) = tensor ∈ keys(tn.tensormap)
 Base.in(index::Symbol, tn::TensorNetwork) = index ∈ keys(tn.indexmap)
 
+function groupinds!(tn::AbstractTensorNetwork, i)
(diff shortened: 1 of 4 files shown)


starsfordummies (approved):
\o/

src/TensorNetwork.jl
```diff
@@ -644,6 +647,38 @@ This method is equivalent to `tensor ∈ tensors(tn)` code, but it's faster on l
 Base.in(tensor::Tensor, tn::TensorNetwork) = tensor ∈ keys(tn.tensormap)
 Base.in(index::Symbol, tn::TensorNetwork) = index ∈ keys(tn.indexmap)
 
+"""
+    groupinds!(tn::AbstractTensorNetwork, i::Symbol)
+
+Group indices parallel to `i` and reshape the tensors to reflect it.
```
jofrevalles, line 653:
```suggestion
Group indices parallel to `i` and reshape the tensors accordingly.
```

jofrevalles (changes requested):

jofrevalles:
Shouldn't we indicate which indices should the function group? Like `:i` and `:j`, and not only those who are "parallel to" `:i`?

mofeing:
I see your point, but we are only going to "group" indices that are parallel; grouping indices that are not parallel involves performing contractions of the involved tensors, which we don't want to force.

PR #208, opened 2024-09-23
Replace OMEinsum with TensorOperations

mofeing:
After some benchmarking, I've found out that OMEinsum is terribly slow and would like to see how TensorOperations performs compared to it. This PR replaces OMEinsum with TensorOperations. If it improves runtime, I will replace our backend for it.

@lkdvos @Jutho I'm running into some edge cases that are managed by OMEinsum but not TensorOperations. Would you mind lending me a hand?


lkdvos:
I'm obviously also curious about the benchmarks, and would gladly help out a bit. This being said, I think one of the main differences between OMEinsumeinsum and Tensoroperations implementation is that TensorOperations really requires strict Einstein summation, every index must appear exactly twice: either once in the input and once in the output, or twice in the inputs such that it is contracted. From a quick first glance, these are the errors you are having?

I am not sure if this is easily resolved, both the contraction order determination and the elementary operations make use of this assumption. For instance, I think the hyper indices would require a major rewrite as we really have that contractions happen pairwise.

I'm not familiar enough with OMEinsum, but maybe there is a way to only dispatch the implementation of certain pairwise contractions to TensorOperations?

lkdvos:
I don't have anything against supporting that, but note that none of these cases are really true Einstein summation things, hence they don't currently work.
I also have to say that I don't have too much time myself to invest into this right now, and at least for us there is not too many reasons to add this: our `ncon` and `@tensor` implementation will not generate calls to this and error earlier. Additionally, within the context of symmetric tensors and TensorKit, none of these cases are very well-defined when considering charge conservation.
This being said, if you would find a nice way to add this, I am not opposed to supporting that for Arrays

mofeing:
I'm thinking that we should have sth like a "EinsumInterface.jl", similar to what "DifferentiationInterface.jl" does for AD engines.

lkdvos:
Maybe, but it's hard to cleanly define what you want to include. TensorOperations does have an interface and backend selection mechanisms, so you could plug in whatever implementation you want quite easily, but limits itself to simple einsum expressions. There is definitely a case to be made for hyperindices, summed indices, but at that point you might also want to consider fixed indices, and maybe you don't want to have just contractions with conjugations, but also allow 'abs' or 'max', and have different reduction operations etc. It's not that obvious where to draw the line, and any line seems as arbitrary as any other line, making it hard to establish a good interface...

mofeing:
@lkdvos I've been thinking how to circunvent these issues and what can we do from Tenet's side. so for each case...:

- case 1: i can just detect and call `sum` on it
- case 2: i can just detect and call `view` or `view` and `collect` on it
- case 4: i can just detect, reshape and call `broadcast(*)`

for case 3, it's a bit harder but i believe what we can do is just project `k`, call `tensorcontract!` for each projection of `k` and stack them.

maybe you want to reconsider support for case 3, not for the explanation i gave before, but because it's equivalent to **batch tensor contraction**, and BLAS implementations are starting to give good support for batch GEMM (but i saw @jutho's answers to issues asking for batch tensor contract).

lkdvos:
I think the main issue will be that you cannot intercept the macro without fully rewriting it, so while your functions will work, the actual `@tensor` calls will not be able to support these edge cases. If you are okay with that, I guess that's not too bad?

lkdvos:
That indeed sounds plausible, and probably not that hard to implement?

mofeing:
Rebased on top of #329

PR #211, opened 2024-09-24
Visualize `AbstractTensorNetworks` interactively with ECharts

mofeing:
# To do

- [ ] Refactor into a `draw` function (just like `quimb`)
  - It should be customizable
- [ ] Take defaults from GraphMakie extension

https://github.com/user-attachments/assets/255b4f0e-1301-4cf1-96e4-325a6a85d54e


Project.toml
```diff
@@ -7,7 +7,9 @@ version = "0.8.0-DEV"
 AbstractTrees = "1520ce14-60c1-5f80-bbc7-55ef81b5835c"
 Combinatorics = "861a8166-3701-5b0c-9a16-15d98fcdc6aa"
 DeltaArrays = "10b0fc19-5ccc-4427-889b-d75dd6306188"
+ECharts_jll = "ffd69456-1935-58d2-abba-ba12e8909167"
```
jofrevalles, line 10: Shouldn't this be an extension? Since we have the visualization libraries `GraphMakie` and `Makie` as an extension too.

mofeing: I see your point but in this case no. The reason is that package extensions are there to avoid precompile and load time overheads.

In this case, ECharts is a JavaScript library (so only gets loaded on the web browser/notebook/VSCode, it doesn't get loaded to the Julia session). Furthermore, JLL packages don't load anything to memory... they are just wrappers over some "binary" files stored on disk. And the minified file just weights 1MiB...

mofeing:
closed in favor of #261

PR #215, opened 2024-10-08
Implement Eigendecomposition

mofeing:
This PR implements `LinearAlgebra.eigen` (and related methods, `eigvals` and `eigvecs`) for `Tensor`, and `LinearAlgebra.eigen!` for `AbstractTensorNetwork`.

The design of this implementation is somewhat more similar to how Julia implements factorizations: there is this new `TensorEigen` object which `eigen(::Tensor)` returns. We could have used `LinearAlgebra.Eigen` (which is the return type of `eigen`), but it doesn't store the right indices required to construct $U\^{-1}$.

The following destructuring syntax is supported:
```julia
A = Tensor(normalize!(rand(2,2)), [:i, :j])

Λ, U = eigen(A)
(; U⁻¹) = eigen(A)
(; Uinv) = eigen(A)
(; U, Λ, U⁻¹) = eigen(A)
```

### DIFF
--- a/src/Numerics.jl
+++ b/src/Numerics.jl
@@ -119,6 +119,72 @@ function factorinds(tensor, left_inds, right_inds)
     return left_inds, right_inds
 end
 
+# TODO is this an `AbstractTensorNetwork`?
+# TODO add fancier `show` method
+struct TensorEigen{T,V,Nᵣ,S<:AbstractVector{V},U<:AbstractArray{T,Nᵣ}} <: Factorization{T}
+    values::Tensor{V,1,S}
+    vectors::Tensor{T,Nᵣ,U}
+    right_inds::Vector{Symbol}
+end
+
+function Base.getproperty(obj::TensorEigen, name::Symbol)
+    if name === :U
+        return obj.vectors
+    elseif name === :Λ
+        return obj.values
+    elseif name ∈ [:Uinv, :U⁻¹]
+        U = reshape(parent(obj.vectors), prod(size(obj.vectors)[1:(end - 1)]), size(obj.vectors)[end])
+        Uinv = inv(U)
+        return Tensor(Uinv, [only(inds(obj.values)), obj.right_inds...])
+    end
+    return getfield(obj, name)
+end
+
+function Base.inv(F::TensorEigen)
+    U = reshape(parent(F.vectors), prod(size(F.vectors)[1:(end - 1)]), size(F.vectors)[end])
+    return Tensor(U * inv(Diagonal(F.values)) / U, [F.left_inds..., F.right_inds...])
+end
+LinearAlgebra.det(x::TensorEigen) = prod(x.values)
+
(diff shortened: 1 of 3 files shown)


jofrevalles:
Okay thanks! Can you add tests for this? See our test for [`KrylovKit.eigsolve`](https://github.com/bsc-quantic/Tenet.jl/blob/master/test/integration/KrylovKit_test.jl), which may be very similar.

Also, we should make that both `KrylovKit.eigsolve` and this function you added should have similar syntax, so if we agree on this `TensorEigen` thing, we should add it to our `KrylovKit` extension.

jofrevalles (commented):
Okay thanks! Can you add tests for this? See our test for [`KrylovKit.eigsolve`](https://github.com/bsc-quantic/Tenet.jl/blob/master/test/integration/KrylovKit_test.jl), which may be very similar.

Also, we should make that both `KrylovKit.eigsolve` and this function you added should have similar syntax, so if we agree on this `TensorEigen` thing, we should add it to our `KrylovKit` extension.

mofeing:
> Okay thanks! Can you add tests for this? See our test for [`KrylovKit.eigsolve`](https://github.com/bsc-quantic/Tenet.jl/blob/master/test/integration/KrylovKit_test.jl?rgh-link-date=2024-10-08T08%3A17%3A38Z), which may be very similar.

Sure

> Also, we should make that both `KrylovKit.eigsolve` and this function you added should have similar syntax, so if we agree on this `TensorEigen` thing, we should add it to our `KrylovKit` extension.

mmm although it might be a good idea, I see that `Krylovkit.eigsolve` doesn't return a `Factorization` object like `LinearAlgebra.eigen` so I would prefer to leave that for another PR.

src/Numerics.jl
```diff
@@ -119,6 +119,73 @@ function factorinds(tensor, left_inds, right_inds)
...
+    left_inds = inds(F.vectors)[1:(end - 1)]
+    return Tensor(U * inv(Diagonal(F.values)) / U, [left_inds..., F.right_inds...])
+end
+LinearAlgebra.det(x::TensorEigen) = prod(x.values)
+
+Base.iterate(x::TensorEigen) = (x.values, :vectors)
+Base.iterate(x::TensorEigen, state) = state == :vectors ? (x.vectors, nothing) : nothing
+
+LinearAlgebra.eigen(t::Tensor{<:Any,2}; kwargs...) = @invoke eigen(t::Tensor; left_inds=(first(inds(t)),), kwargs...)
+function LinearAlgebra.eigen(tensor::Tensor; left_inds=(), right_inds=(), virtualind=Symbol(uuid4()), kwargs...)
```
jofrevalles, line 154: Nice! Can you add a `docstring`? :)

mofeing (commented):
> Nice! Can you add a `docstring`? :) 

Done!

PR #218, opened 2024-10-09
Implement an MPS method initializing the tensors to identity (copy-tensors)

### LINKED ISSUE #165: Implement identity initialization of `Chain`
The method is described in https://arxiv.org/pdf/2203.03366.pdf

Todorbsc:
Closes #165

### DIFF
--- a/src/Ansatz/MPS.jl
+++ b/src/Ansatz/MPS.jl
@@ -77,14 +77,14 @@ function Base.identity(::Type{MPS}, n; physdim=2, maxdim=physdim^(n ÷ 2), order
         throw(ArgumentError("order must be a permutation of $(String.(defaultorder(MPS)))"))
 
     # Create bond dimensions until the middle of the MPS considering maxdim
-    virtualdims = physdim .^ collect(1:n ÷ 2)
+    virtualdims = physdim .^ collect(1:(n ÷ 2))
     virtualdims = ifelse.((virtualdims .> maxdim), maxdim, virtualdims)
     # Complete the bond dimensions of the other half of the MPS
-    virtualdims = vcat(virtualdims, reverse(n % 2 == 1 ? virtualdims : virtualdims[1:end-1]))
+    virtualdims = vcat(virtualdims, reverse(n % 2 == 1 ? virtualdims : virtualdims[1:(end - 1)]))
 
     # Create each site dimensions in default order (:o, :l, :r)
     arraysdims = [[physdim, virtualdims[1]]]
-    append!(arraysdims, [[physdim, virtualdims[i], virtualdims[i+1]] for i in 1:(length(virtualdims)-1)])
+    append!(arraysdims, [[physdim, virtualdims[i], virtualdims[i + 1]] for i in 1:(length(virtualdims) - 1)])
     push!(arraysdims, [physdim, virtualdims[end]])
 
     # Create the MPS with copy-tensors according to the tensors dimensions
--- a/src/Ansatz/MPS.jl
+++ b/src/Ansatz/MPS.jl
@@ -80,6 +80,7 @@ function Base.identity(::Type{MPS}, n::Integer; physdim=2, maxdim=physdim^(n ÷ 
 
     # Create bond dimensions until the middle of the MPS considering maxdim
     virtualdims = min.(maxdim, physdim .^ (1:(n ÷ 2)))
+
     # Complete the bond dimensions of the other half of the MPS
     virtualdims = vcat(virtualdims, virtualdims[(isodd(n) ? end : end - 1):-1:1])
 
--- a/src/Ansatz/MPS.jl
+++ b/src/Ansatz/MPS.jl
@@ -62,12 +62,17 @@ function MPS(arrays::Vector{<:AbstractArray}; order=defaultorder(MPS))
     return MPS(ansatz, NonCanonical())
 end
 
-MPS(arraysdims::NTuple{N,<:Vector{Int}}; order=defaultorder(MPS)) where {N} = MPS(collect(arraysdims); order=order)
-MPS(arraysdims::Vector{<:Tuple{Vararg{Int}}}; order=defaultorder(MPS)) = MPS(collect.(arraysdims); order=order)
-function MPS(arraysdims::NTuple{N,Tuple{Vararg{Int}}}; order=defaultorder(MPS)) where {N}
-    return MPS(collect(collect.(arraysdims)); order=order)
-end
-function MPS(arraysdims::Vector{<:Vector{Int}}; order=defaultorder(MPS))
+"""
+    Base.identity(::Type{MPS}, arraysdims; order=defaultorder(MPS))
+
+Returns an [`MPS`](@ref) whose tensors are initialized to the identity with the specified 
+dimensions from `arraysdims`.
+
+# Keyword Arguments
+
+  - `order` Tensors' indices order. Default: output - left - right (:o, :l, :r).
+"""
+function Base.identity(::Type{MPS}, arraysdims; order=defaultorder(MPS))
     @assert length(arraysdims[1]) == 2 "First array must have 2 dimensions"
     @assert all(==(3) ∘ length, arraysdims[2:(end - 1)]) "All arrays must have 3 dimensions"
     @assert length(arraysdims[end]) == 2 "Last array must have 2 dimensions"
--- a/src/Ansatz/MPS.jl
+++ b/src/Ansatz/MPS.jl
@@ -65,7 +65,7 @@ end
 """
     Base.identity(::Type{MPS}, arraysdims; order=defaultorder(MPS))
 
-Returns an [`MPS`](@ref) whose tensors are initialized to the identity with the specified 
+Returns an [`MPS`](@ref) whose tensors are initialized to the identity with the specified
 dimensions from `arraysdims`.
 
 # Keyword Arguments
--- a/src/Ansatz/MPS.jl
+++ b/src/Ansatz/MPS.jl
@@ -63,16 +63,18 @@ function MPS(arrays::Vector{<:AbstractArray}; order=defaultorder(MPS))
 end
 
 """
-    Base.identity(::Type{MPS}, arraysdims; order=defaultorder(MPS))
+    Base.identity(::Type{MPS}, n::Integer; physdim=2, maxdim=physdim^(n ÷ 2), order=MPS.defaultorder(MPS))
 
-Returns an [`MPS`](@ref) whose tensors are initialized to the identity with the specified
-dimensions from `arraysdims`.
+Returns an [`MPS`](@ref) of `n` sites whose tensors are initialized to the identity attending to the
+physical dimension `physdim` and the maximum dimension `maxdim`.
 
 # Keyword Arguments
 
-  - `order` Tensors' indices order. Default: output - left - right (:o, :l, :r).
+  - `physdim` The physical or output dimension of each site. Default is 2
+  - `maxdim` The maximum bond dimension. Default is `physdim^(n ÷ 2)`.
+  - `order` Tensors' indices order. Default: output - left - right `(:o, :l, :r)`.
 """
-function Base.identity(::Type{MPS}, n; physdim=2, maxdim=physdim^(n ÷ 2), order=defaultorder(MPS))
+function Base.identity(::Type{MPS}, n::Integer; physdim=2, maxdim=physdim^(n ÷ 2), order=defaultorder(MPS))
     issetequal(order, defaultorder(MPS)) ||
         throw(ArgumentError("order must be a permutation of $(String.(defaultorder(MPS)))"))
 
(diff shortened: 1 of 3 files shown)


test/MPS_test.jl
```diff
@@ -26,6 +26,28 @@
...
+        @test t1[1, 1] == t1[2, 2] == 1
+        @test sum(t1) == 2
+        t2 = tensors(ψ; at=site"2")
+        @test t2[1, 1, 1] == t2[2, 2, 2] == t2[3, 3, 3] == 1
+        @test sum(t2) == 3
+        t3 = tensors(ψ; at=site"3")
+        @test t3[1, 1] == t3[2, 2] == 1
+        @test sum(t3) == 2
+    end
+
```
mofeing, line 50: Wrap this into a `@testset`

src/Ansatz/MPS.jl
```diff
@@ -62,6 +62,30 @@ function MPS(arrays::Vector{<:AbstractArray}; order=defaultorder(MPS))
     return MPS(ansatz, NonCanonical())
 end
 
+MPS(arraysdims::NTuple{N,<:Vector{Int}}; order=defaultorder(MPS)) where {N} = MPS(collect(arraysdims); order=order)
+MPS(arraysdims::Vector{<:Tuple{Vararg{Int}}}; order=defaultorder(MPS)) = MPS(collect.(arraysdims); order=order)
+function MPS(arraysdims::NTuple{N,Tuple{Vararg{Int}}}; order=defaultorder(MPS)) where {N}
+    return MPS(collect(collect.(arraysdims)); order=order)
+end
```
mofeing, line 69: I would currently just delete these methods for now...
```suggestion
```

src/Ansatz/MPS.jl
```diff
@@ -62,6 +62,30 @@ function MPS(arrays::Vector{<:AbstractArray}; order=defaultorder(MPS))
...
+function MPS(arraysdims::Vector{<:Vector{Int}}; order=defaultorder(MPS))
```
mofeing, line 70: I would use `Base.identity`, not a constructor since you're not passing arrays.
```suggestion
function Base.identity(::Type{MPS}, arraysdims; order=defaultorder(MPS))
```

mofeing (changes requested):
Thanks! But could you clarify a lil bit how it works?

src/Ansatz/MPS.jl
```diff
@@ -62,6 +62,30 @@ function MPS(arrays::Vector{<:AbstractArray}; order=defaultorder(MPS))
...
+            mindim = minimum(arrdims)
+            arr = zeros(ComplexF64, arrdims...)
+            deltas = ntuple(x -> ntuple(_ -> x, length(arrdims)), mindim)
+            broadcast(delta -> arr[delta...] = 1.0, deltas)
+            arr
+        end;
+        order=order,
+    )
+end
+
```
jofrevalles, line 88: I would not create another constructor for `MPS`, since that may be misleading to the user who wants to directly create an `MPS` with some specific arrays. Instead, you could extend the function `ones` or one similiar, to make that when we call `ones(MPS, arraydims ... )` we get this kind of `MPS`. Something similar to the things we do with `rand`.

jofrevalles (commented):
Thanks @Todorbsc

jofrevalles:
Also note that, as we discussed, we should add a `docstring` to every new function we make, since we lack a lot of documentation. :)

test/MPS_test.jl
Todorbsc: Done

src/Ansatz/MPS.jl
Todorbsc: I removed them

src/Ansatz/MPS.jl
Todorbsc: This was one possibility along with the one @jofrevalles suggested below of using `ones`. I replaced the constructor by `identity` as I think it fits better than `ones`.

src/Ansatz/MPS.jl
Todorbsc: As I said above, I agree it is better to not implement another constructor. The option that fits the best for me is `Base.identity`, check it out :)

Todorbsc:
> Thanks! But could you clarify a lil bit how it works?

Sure! The point is to create an identity `MPS` where the tensors will have some user-provided dimensions (`arraysdims`). Thus, what `Base.identity(::Type{MPS}, arraysdims; order)` would do is create zero-initialized tensors with the specified dimensions and then assign to the unit those tensor elements that have the same index in all dimensions.
For example: In a tensor of shape `(3,4,5)`, the constraint is the lowest dimension of the tensor, i.e. `3` allowing the assignment to the unit the tensors elements `(1,1,1)`, `(2,2,2)` and `(3,3,3)`.

src/Ansatz/MPS.jl
mofeing: Yeah, I think `ones` can be misleading since it could mean you're initializing the MPS to a $\ket{11\dots 1}$ state.

I think `identity` is better because contracting against another "identity" MPS gives you 1, all the tensors are just COPY-tensors and "identity initialization" is how is known in some packages (it's also known as `eye`-initialization but `eye` is how the identity is known on those languages).

mofeing:
> Sure! The point is to create an identity `MPS` where the tensors will have some user-provided dimensions (`arraysdims`). Thus, what `Base.identity(::Type{MPS}, arraysdims; order)` would do is create zero-initialized tensors with the specified dimensions and then assign to the unit those tensor elements that have the same index in all of its dimensions. For example: In a tensor of shape `(3,4,5)`, the constraint is the lowest dimension of the tensor, i.e. `3`, doing the unit assignment to the tensors elements `(1,1,1)`, `(2,2,2)` and `(3,3,3)`.

Thanks for clarifying but I think have custom dimensions is not so interesting and can be tedious for the user to generate all the tuples...

IMO it would be best to accept a signature like
```julia
Base.identity(::Type{MPS}, n; maxdim=2^(n ÷ 2), physdim=2)
```
where `n` is the number of sites and `maxdim` is the $\chi$ and is the maximum bond dimensions. The size of the virtual dims would then increase exponentially fast until arriving to `maxdim`.

Todorbsc:
> ```julia
> maxdim=2^(n ÷ 2), physdim=2
> ```

@mofeing What happens in the case there $\chi$ (`maxdim`) is less than the middle of the `MPS`?
For instance, a 7-site `MPS` with `maxdim=4`: the 1st virtual index would be 2-dimensional and the 2nd one would be 4-dimensional (where the `maxdim` is reached). Would the 3rd one (reaching to the half of the chain) also be 4-dimensional?
This would give the following bond dimensions [2, 4, 4, 4, 4, 2]. Is this right?

mofeing:
> @mofeing What happens in the case where χ (`maxdim`) is reached before the middle of the `MPS`? For instance, a 7-site `MPS` with `maxdim=4`: the 1st virtual index would be 2-dimensional and the 2nd one would be 4-dimensional (where the `maxdim` is reached). Would the 3rd one (reaching to the half of the chain) also be 4-dimensional? This would give the following bond dimensions `[2, 4, 4, 4, 4, 2]`. Is this right?

the order or num dims of the tensors is not affected. just the size of the virtual indices is upper bounded. so yeah, `[2, 4, 4, 4, 4, 2]` is correct.

you can use this code in here to compute the virtual index sizes for an open MPS
https://github.com/bsc-quantic/Tenet.jl/blob/4c16d2f9219aa214afa52252030aad06a191b120/src/Ansatz/Chain.jl#L233-L239

Todorbsc:
> > @mofeing What happens in the case where χ (`maxdim`) is reached before the middle of the `MPS`? For instance, a 7-site `MPS` with `maxdim=4`: the 1st virtual index would be 2-dimensional and the 2nd one would be 4-dimensional (where the `maxdim` is reached). Would the 3rd one (reaching to the half of the chain) also be 4-dimensional? This would give the following bond dimensions `[2, 4, 4, 4, 4, 2]`. Is this right?
> 
> the order or num dims of the tensors is not affected. just the size of the virtual indices is upper bounded. so yeah, `[2, 4, 4, 4, 4, 2]` is correct.
> 
> you can use this code in here to compute the virtual index sizes for an open MPS
> 
> https://github.com/bsc-quantic/Tenet.jl/blob/4c16d2f9219aa214afa52252030aad06a191b120/src/Ansatz/Chain.jl#L233-L239

Perfect, thanks! 
Actually, I had written my own method for the bond dimensions before reading this, check it out and I can change it if you wish.

test/MPS_test.jl
```diff
@@ -26,6 +26,49 @@
...
+            # - Test middle tensors (3D) equal identity
+            diagonal_3D = [fill(i, 3) for i in 1:physdim]
+            @test all(tns -> all(delta -> tns[delta...] == 1, diagonal_3D), alltns[2:(end - 1)])
+            @test all(tns -> sum(tns) == physdim, alltns[2:(end - 1)])
+
+            # Test whether the contraction gives the identity
+            contracted_ψ = contract(ψ)
+            diagonal_nsitesD = [fill(i, nsites) for i in 1:physdim]
+            @test all(delta -> contracted_ψ[delta...] == 1, diagonal_nsitesD)
+            @test sum(contracted_ψ) == physdim
```
jofrevalles, line 68:
```suggestion
        for (nsites, physdim, expected_tensorsizes, maxdim) in
            zip(nsites_cases, physdim_cases, expected_tensorsizes_cases, maxdim_cases)
            ψ = if isnothing(maxdim)
                identity(MPS, nsites; physdim=physdim)
            else
                identity(MPS, nsites; physdim=physdim, maxdim=maxdim)
            end
            
            # Test the tensor dimensions
            obtained_tensorsizes = size.(tensors(ψ))
            @test obtained_tensorsizes == expected_tensorsizes

            # Test whether all tensors are the identity
            alltns = tensors(ψ)
            
            # - Test extreme tensors (2D) equal identity
            diagonal_2D = [fill(i, 2) for i in 1:physdim]
            @test all(delta -> alltns[1][delta...] == 1, diagonal_2D)
            @test sum(alltns[1]) == physdim
            @test all(delta -> alltns[end][delta...] == 1, diagonal_2D)
            @test sum(alltns[end]) == physdim
            
            # - Test bulk tensors (3D) equal identity
            diagonal_3D = [fill(i, 3) for i in 1:physdim]
            @test all(tns -> all(delta -> tns[delta...] == 1, diagonal_3D), alltns[2:(end - 1)])
            @test all(tns -> sum(tns) == physdim, alltns[2:(end - 1)])

            # Test whether the contraction gives the identity
            contracted_ψ = contract(ψ)
            diagonal_nsitesD = [fill(i, nsites) for i in 1:physdim]
            @test all(delta -> contracted_ψ[delta...] == 1, diagonal_nsitesD)
            @test sum(contracted_ψ) == physdim
```

jofrevalles (changes requested):
Nice! I would add some empty lines to make the testset clearer, but this is ready to merge soon.

src/Ansatz/MPS.jl
```diff
@@ -62,6 +62,44 @@ function MPS(arrays::Vector{<:AbstractArray}; order=defaultorder(MPS))
...
+  - `maxdim` The maximum bond dimension. Default is `physdim^(n ÷ 2)`.
+  - `order` Tensors' indices order. Default: output - left - right `(:o, :l, :r)`.
+"""
+function Base.identity(::Type{MPS}, n::Integer; physdim=2, maxdim=physdim^(n ÷ 2), order=defaultorder(MPS))
+    issetequal(order, defaultorder(MPS)) ||
+        throw(ArgumentError("order must be a permutation of $(String.(defaultorder(MPS)))"))
+
+    # Create bond dimensions until the middle of the MPS considering maxdim
+    virtualdims = min.(maxdim, physdim .^ (1:(n ÷ 2)))
+    # Complete the bond dimensions of the other half of the MPS
```
mofeing, line 83:
```suggestion

    # Complete the bond dimensions of the other half of the MPS
```

mofeing (changes requested):

src/Ansatz/MPS.jl
```diff
@@ -62,6 +62,42 @@ function MPS(arrays::Vector{<:AbstractArray}; order=defaultorder(MPS))
...
+    Base.identity(::Type{MPS}, n::Integer; physdim=2, maxdim=physdim^(n ÷ 2))
+
+Returns an [`MPS`](@ref) of `n` sites whose tensors are initialized to the identity
+in default order `(:o, :l, :r)` attending to the physical dimension `physdim` and
+the maximum bond dimension `maxdim`.
+
+# Keyword Arguments
+
+  - `physdim` The physical or output dimension of each site. Default is 2
+  - `maxdim` The maximum bond dimension. Default is `physdim^(n ÷ 2)`.
```
mofeing, line 75:
```suggestion
Returns an [`MPS`](@ref) of `n` sites whose tensors are initialized to COPY-tensors.

# Keyword Arguments

  - `physdim` The physical or output dimension of each site. Defaults to 2.
  - `maxdim` The maximum bond dimension. Defaults to `physdim^(n ÷ 2)`.
```

mofeing (approved):

PR #223, opened 2024-10-26
Refactor `Ansatz` to be a concrete type with lattice information

mofeing:
As discussed offline, we decided to slice #204 into smaller PRs. This first PR makes `Ansatz` a concrete type with lattice information contained in a graph.

The important thing you should review is the API, not the details; we can change the graph type in the future but we want the API to be stable.

# Changelog
- Introduced `Lattice` graph type
  - It's basically a `Graphs.SimpleGraph` with a bijective mapping between vertices and `Site`s
  - It adds `Graphs` ~and `MetaGraphsNext`~ as dependenc~ies~y
- Make `Ansatz` a concrete type in the type hierarchy
  - Its abstract counterpart is `AbstractAnsatz`
- Delete `Chain`, `Product`, `Grid` and `Dense`
  - These will be re-added in subsequent PRs, adapted to the new `Ansatz` layer
- Make `Form` the canonical form information trait
- Refactor some `evolve!` methods based on simple-update to `simple_update!` and generalize it for any `AbstractAnsatz`
- Added `@testset_skip` macro to implement @jofrevalles's idea of skipping whole testsets temporarily

# Notes
- ~Currently there is a bug in Julia makes Enzyme crash on an `Ansatz` due to its field `lattice` being abstract https://github.com/EnzymeAD/Enzyme.jl/issues/1923~ Should no longer be necessary because we replaced `MetaGraph` for `Lattice`
  - ~It's fixed in Julia 1.11.0 and should be fixed in Julia ~1.10.6~ 1.10.7~
  - ~I temporarily fixed it by fixing `Ansatz` to just 1D connectivity (i.e. `Site{1}`) but I'm removing it now -> For the time being you will need Julia 1.11 if you want to diff an `Ansatz`~
- ~There are other bug fixes (like `Base.in` or `GraphMakie.graphplot!` on `AbstractTensorNetwork`) that I'm gonna remove now and move to other PRs~ Already pushed to master
- Might change `Lattice` to `Topology` in the future, because it's more accurate if we're not using the partial ordering property of lattices
- In the future, we might change the subjacent graph type in `Lattice` for custom ones to accommodate clustered sites (i.e. a tensor containing 2+ sites) or hypergraphs


src/Tenet.jl
```diff
@@ -26,24 +26,9 @@ include("Ansatz/Ansatz.jl")
 export Ansatz
 export socket, Scalar, State, Operator
 export boundary, Open, Periodic
+export form
```
jofrevalles, line 30:
```suggestion
```

jofrevalles (changes requested):

src/Tenet.jl
mofeing: why not export it?

jofrevalles:
@mofeing @starsfordummies I think that when we talked about splitting the large PR into small ones, we meant that this PR would only contain the new `Ansatz` struct, and I see here that you pushed the commits regarding the canonical form, which IMO they should be in another PR. In this way, we can refactor the tests that are failing (like `Chain`, `Dense`, ...) in another PR.

mofeing:
> @mofeing @starsfordummies I think that when we talked about splitting the large PR into small ones, we meant that this PR would only contain the new `Ansatz` struct, and I see here that you pushed the commits regarding the canonical form, which IMO they should be in another PR. In this way, we can refactor the tests that are failing (like `Chain`, `Dense`, ...) in another PR.

Sorry, but I don't agree. The `Form` trait is part of `Ansatz`, even if it doesn't have a field for that.

The other ansatzes will be reimplemented in subsequent PRs.

test/integration/YaoBlocks_test.jl
```diff
@@ -8,35 +8,35 @@
     @test issetequal(sites(tn), [site"1", site"2", site"1'", site"2'"])
     @test Tenet.ntensors(tn) == 2
 
-    @testset "GHZ Circuit" begin
-        circuit_GHZ = chain(n_qubits, put(1 => Yao.H), Yao.control(1, 2 => Yao.X), Yao.control(2, 3 => Yao.X))
+    # @testset "GHZ Circuit" begin
```
jofrevalles, line 11:
```suggestion
    @test_skip begin #@testset "GHZ Circuit"
```

test/Product_test.jl
```diff
@@ -1,33 +1,33 @@
-@testset "Product ansatz" begin
-    using LinearAlgebra
+# @testset "Product ansatz" begin
```
jofrevalles, line 1:
```suggestion
@test_skip begin # @testset "Product ansatz"
```

test/Ansatz_test.jl
```diff
@@ -0,0 +1,2 @@
+@testset "Ansatz" begin
```
jofrevalles, line 1: Add tests.

src/Ansatz/Ansatz.jl
```diff
@@ -36,43 +81,280 @@ Base.show(io::IO, tn::A) where {A<:Ansatz} = summary(io, tn)
...
+
+Truncate the dimension of the virtual `bond`` of an [`Ansatz`](@ref) Tensor Network by keeping only the `maxdim` largest Schmidt coefficients or those larger than`threshold`.
 
-    return tn[only(inds(tensor1) ∩ inds(tensor2))]
+# Notes
+
+  - Either `threshold` or `maxdim` must be provided. If both are provided, `maxdim` is used.
+  - The bond must contain the Schmidt coefficients, i.e. a site canonization must be performed before calling `truncate!`.
+"""
+function truncate!(tn::AbstractAnsatz, bond; threshold=nothing, maxdim=nothing)
```
jofrevalles, line 123: IMO we should have truncate only for `Canonical` or `MixedCanonical` now

jofrevalles (changes requested):
I prefer to have `@test_skip` instead of commenting all the tests.

mofeing:
> I prefer to have `@test_skip` instead of commenting all the tests.

That's a valid request and a good idea. Will change it now and recover the deleted tests as `@test_skip`... The only thing is I'm not sure if it will still try to evaluate the code, which will obviously crash because we're breaking the API. So maybe `@test_broken` but doesn't matter which.

test/Ansatz_test.jl
mofeing: I'm on it.

src/Ansatz/Ansatz.jl
mofeing: I don't agree. Canonizing a QTN can be costly and truncating a non-canonical TN is correctly defined. It is true that local truncation is far from optimal, but there are cases where you don't have an alternative or it's good enough (e.g. the graphene paper used local truncation without canonization and worked good enough).

test/integration/YaoBlocks_test.jl
mofeing: I added a new `@testset_skip` macro for this.

test/runtests.jl
```diff
@@ -10,8 +12,10 @@ using OMEinsum
     include("Transformations_test.jl")
     include("Site_test.jl")
     include("Quantum_test.jl")
-    include("Product_test.jl")
-    include("Chain_test.jl")
+    include("Lattice_test.jl")
+    include("Ansatz_test.jl")
+    # include("Product_test.jl")
+    # include("Chain_test.jl")
```
jofrevalles, line 18:
```suggestion
    include("Product_test.jl")
    include("Chain_test.jl")
```
If we have `testskip` inside we don't need to comment out this, right?

src/Ansatz.jl
```diff
@@ -0,0 +1,485 @@
...
+Truncate the dimension of the virtual `bond`` of an [`Ansatz`](@ref) Tensor Network by keeping only the `maxdim` largest Schmidt coefficients or those larger than`threshold`.
+
+# Notes
+
+  - Either `threshold` or `maxdim` must be provided. If both are provided, `maxdim` is used.
+  - The bond must contain the Schmidt coefficients, i.e. a site canonization must be performed before calling `truncate!`.
+"""
+function truncate!(tn::AbstractAnsatz, bond; threshold=nothing, maxdim=nothing)
+    @assert isnothing(maxdim) ⊻ isnothing(threshold) "Either `threshold` or `maxdim` must be provided"
+
```
jofrevalles, line 217:
```suggestion
    @assert form(ψ) != NonCanonical() "A `NonCanonical` tensor network can not be truncated"

```

jofrevalles (changes requested):

src/Ansatz.jl
mofeing: Again, I don't agree with this affirmation. You CAN truncate even if it's far from optimal, and in many cases it works.

jofrevalles:
If we fix the documentation I guess we can merge

mofeing:
> If we fix the documentation I guess we can merge

Docs are basically broken. We need to redo them but after this PR.

jofrevalles (approved):

starsfordummies (approved):
Sergio breaks tenet in all possible ways without even documenting what's broken. Typical commit, looks good to me 👍

arturgs (approved):

PR #227, opened 2024-10-31
Fix `replace!` function when replacing a `Tensor` with itself in `pair` argument

### LINKED ISSUE #225: Incorrect `replace!` of a `Tensor` in a `TensorNetwork` with the same `Tensor`
### Summary
The `replace!` function for a `Tensor` in a `TensorNetwork`, where it is replaced with the same `Tensor`, leads to the deletion of that `Tensor`:
```julia
julia> using Tenet

julia> B = Tensor(rand(Float32, 2, 2), (:k, :j))
2×2 Tensor{Float32, 2, Matrix{Float32}}: ...

julia> replace!(tn, A => (A))
TensorNetwork (#tensors=1, #inds=2)

julia> tensors(tn) .|> inds
1-element Vector{Vector{Symbol}}:
 [:k, :j]
```
I think that there might be a conditional in the `replace!` function that checks whether it's the same `Tensor`, and this may be causing the problem.

jofrevalles:
### Summary
This PR resolves #225 by fixing the problem with `replace!` when the same `Tensor` appears in both positions of the `pair` argument. Previously, since we did not have any specific check for this case, the `replace!` function had a `push!` that did not add the tensor since it was already present, but then, because we had a `delete!` after the `push!`, the `Tensor` got deleted.

We fixed this problem by adding a condition in the `replace!` function that checks if the two tensors are equal, and if they are, the function returns the `TensorNetwork` without changes.

Additionally (and importantly @bsc-quantic/software !), this PR also fixes the commented-out tests for the `replace!` function, and we have extended these tests to cover the specific case that this PR addresses.

### DIFF
--- a/src/TensorNetwork.jl
+++ b/src/TensorNetwork.jl
@@ -438,7 +438,7 @@ function Base.replace!(tn::AbstractTensorNetwork, pair::Pair{<:Tensor,<:Tensor})
     tn = TensorNetwork(tn)
     old_tensor, new_tensor = pair
 
-    old_tensor == new_tensor && return tn
+    old_tensor === new_tensor && return tn
 
     issetequal(inds(new_tensor), inds(old_tensor)) || throw(ArgumentError("replacing tensor indices don't match"))
 
--- a/src/TensorNetwork.jl
+++ b/src/TensorNetwork.jl
@@ -437,6 +437,9 @@ end
 function Base.replace!(tn::AbstractTensorNetwork, pair::Pair{<:Tensor,<:Tensor})
     tn = TensorNetwork(tn)
     old_tensor, new_tensor = pair
+
+    old_tensor == new_tensor && return tn
+
     issetequal(inds(new_tensor), inds(old_tensor)) || throw(ArgumentError("replacing tensor indices don't match"))
 
     push!(tn, new_tensor)
--- a/test/TensorNetwork_test.jl
+++ b/test/TensorNetwork_test.jl
@@ -486,7 +486,7 @@
                 end
             end
 
-            @testset "TensorNetwork with tensors with equal indices" begin
+            @testset "TensorNetwork with tensors of equal indices" begin
                 A = Tensor(rand(2, 2), (:u, :w))
                 B = Tensor(rand(2, 2), (:u, :w))
                 tn = TensorNetwork([A, B])
@@ -503,7 +503,7 @@
                 @test issetequal(tensors(tn), [new_tensor, B])
             end
 
-            @testset "Chain of replacements" begin
+            @testset "Sequence of replacements" begin
                 A = Tensor(zeros(2, 2), (:i, :j))
                 B = Tensor(zeros(2, 2), (:j, :k))
                 C = Tensor(zeros(2, 2), (:k, :l))
@@ -518,7 +518,7 @@
                 @test issetequal(tensors(tn), [new_tensor2, B, C])
             end
 
-            @testset "Same Tensor in pair argument" begin
+            @testset "Replace with itself" begin
                 A = Tensor(rand(2, 2), (:i, :j))
                 B = Tensor(rand(2, 2), (:j, :k))
                 C = Tensor(rand(2, 2), (:k, :l))
--- a/test/TensorNetwork_test.jl
+++ b/test/TensorNetwork_test.jl
@@ -459,60 +459,64 @@
         end
 
         @testset "replace tensors" begin
-            t_ij = Tensor(zeros(2, 2), (:i, :j))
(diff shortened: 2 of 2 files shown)


src/TensorNetwork.jl
```diff
@@ -437,6 +437,9 @@ end
 function Base.replace!(tn::AbstractTensorNetwork, pair::Pair{<:Tensor,<:Tensor})
     tn = TensorNetwork(tn)
     old_tensor, new_tensor = pair
+
+    old_tensor == new_tensor && return tn
```
mofeing, line 441: mmm shouldn't this be `===` to check for egality, not equality?
```suggestion
    old_tensor === new_tensor && return tn
```

test/TensorNetwork_test.jl
```diff
@@ -459,60 +459,75 @@
...
+                for ind in inds(new_tensor)
+                    tensors_with_ind = tn.indexmap[ind]
+                    @test new_tensor ∈ tensors_with_ind
+                    @test !(old_tensor ∈ tensors_with_ind)
+                end
             end
 
-            new_tensor = Tensor(rand(2, 2), (:l, :m))
-            replace!(tn, old_tensor => new_tensor)
+            @testset "TensorNetwork with tensors with equal indices" begin
```
mofeing, line 489:
```suggestion
            @testset "TensorNetwork with tensors of equal indices" begin
```

test/TensorNetwork_test.jl
```diff
@@ -459,60 +459,75 @@
...
+                replace!(tn, A => new_tensor)
 
-            # new_tensor = Tensor(rand(2, 2), (:u, :w))
+                @test issetequal(tensors(tn), [new_tensor, B])
+            end
 
-            # replace!(tn, B => new_tensor)
-            # @test A === tensors(tn)[1]
-            # @test new_tensor === tensors(tn)[2]
+            @testset "Chain of replacements" begin
```
mofeing, line 506:
```suggestion
            @testset "Sequence of replacements" begin
```

test/TensorNetwork_test.jl
```diff
@@ -459,60 +459,75 @@
...
-            # A = Tensor(zeros(2, 2), (:i, :j))
-            # B = Tensor(zeros(2, 2), (:j, :k))
-            # C = Tensor(zeros(2, 2), (:k, :l))
-            # tn = TensorNetwork([A, B, C])
+                replace!(tn, A => new_tensor, new_tensor => new_tensor2)
+                @test issetequal(tensors(tn), [new_tensor2, B, C])
+            end
 
-            # @test_throws ArgumentError replace!(tn, A => B, B => C, C => A)
+            @testset "Same Tensor in pair argument" begin
```
mofeing, line 521: How about...?
```suggestion
            @testset "Replace with itself" begin
```

mofeing (changes requested):

mofeing:
Awesome! I'm leaving one import thing (egality vs equality) and some grammatical fix requests.

src/TensorNetwork.jl
jofrevalles: No! Since for example, now it failed when we did `replace!(tn, A => conj(A))`, when `A` is a `Tensor` with a parent `Array` that are `Real`.

mofeing: Exactly! In that case, `A === conj(A)`!

jofrevalles: `conj(A)` returns a new object with is equal to `A`. It would be like you say if it is `conj!` (not `conj`).

mofeing: You're right with this last point, but the bug in #225 is when it's replaced with the *same* tensor (i.e. `old_tensor === new_tensor`).

I try to avoid `==` if posible because `==` iterates over the whole arrays and checks for equality element-wise, while `===` just checks the `objectid`s and you're done. Thus it avoids the overhead of iterating over everything, which can be costly on large arrays (specially when we start using distributed arrays).

jofrevalles: > but the bug in #225 is when it's replaced with the _same_ tensor (i.e. `old_tensor === new_tensor`)

Yes, but I tried `replace!(tn, A => conj(A))` and we got also the same error. I think this is because if you do `conj(A)` you still get the same `objectid`, which for me this is strange:
```julia
julia> objectid(A)
0xc7798e05eba779fe

julia> objectid(conj(A))
0xc7798e05eba779fe
```

But whatever, okay! If that is the case I agree that we can have `===`, I will push that commit now.

jofrevalles:
@mofeing I am merging this PR on monday. Let's enjoy the weekend :)

PR #229, opened 2024-11-05
Fix `conj` to use `deepcopy` instead of `copy`

jofrevalles:
### Summary
In this PR, we change `Base.conj` function to use `deepcopy` instead of `copy` to prevent errors when manipulating tensor networks within an `@unsafe_region` macro block.

The regular `copy` leads to problem since it calls `TensorNetworks(tensors(tn))`, which throws an error since there are inconsistent indices. Before the recent commit a2419673fbb1b6c7f1cf86f5b771efb79a7dd8e5 that refactored the `conj` and `adjoint` functions, we also had a `deepcopy` inside `conj` which prevent this problem.

### Example
With this PR, we can do something like the following code without any problem:
```julia
julia> using Tenet

julia> tn = TensorNetwork([
           Tensor(ones(2, 2), [:a, :b]),
           Tensor(ones(2, 2), [:b, :c])
           ])

julia> Tenet.@unsafe_region tn begin
           c = Tensor(ones(3, 2), [:c, :d])

           push!(tn, c)
           conj(tn)
           pop!(tn, c)
       end
3×2 Tensor{Float64, 2, Matrix{Float64}}:
 1.0  1.0
 1.0  1.0
 1.0  1.0
```

### DIFF
--- a/src/TensorNetwork.jl
+++ b/src/TensorNetwork.jl
@@ -94,7 +94,7 @@ Base.eltype(tn::AbstractTensorNetwork) = promote_type(eltype.(tensors(tn))...)
 Return a copy of the [`AbstractTensorNetwork`](@ref) with all tensors conjugated.
 """
 function Base.conj(tn::AbstractTensorNetwork)
-    tn = copy(tn)
+    tn = deepcopy(tn)
     replace!(tn, tensors(tn) .=> conj.(tensors(tn)))
     return tn
 end


jofrevalles:
@mofeing Okay I got another idea, we can keep the `copy` but add modify it like this:
```
function Base.copy(tn::TensorNetwork)
    if is_unsafe_region[]
        new_tn = TensorNetwork([])
        for tensor in tensors(tn)
            push!(new_tn, copy(tensor))
        end

        return new_tn
    else
        return TensorNetwork(tensors(tn))
    end
end
```

Since `push!` accounts for the `unsafe_region`. Or we can directly add that conditional inside the `TensorNetwork` constructor.

jofrevalles:
> Okay, I like this idea more. The one about adding the check inside `TensorNetwork` constructor.

@mofeing If we do this, then we should agree to change the macro to my second option [described](https://github.com/bsc-quantic/Tenet.jl/issues/230#:~:text=Change%20macro%20to,any%20tensor%20networks.) in issue #230. For me, it makes sense.

mofeing:
Okay, I like the idea but we have to discuss consequences (e.g. we might have problems when multithreading). Add it to the agenda for today's meeting please.

jofrevalles:
@mofeing Since now we fixed the problems with `copy` inside some `@unsafe_region` in PR #233, now we don't need this PR.

mofeing:
Ah really? Nice!

PR #233, opened 2024-11-08
Refactor `@unsafe_region` macro

### LINKED ISSUE #230: `@unsafe_region` applies to all `TensorNetwork`s within the block, not just the one passed in the argument
### Summary
I've noticed that the `@unsafe_region` macro, when applied to a specific `TensorNetwork`, actually affects all tensor networks within its block, even those not specified. This behavior is unexpected and can lead to potential silent errors.

### Example
```julia
julia> using Tenet

julia> tn = TensorNetwork([
           Tensor(ones(2, 2), [:a, :b]),
           Tensor(ones(2, 2), [:b, :c])
       ])

julia> tn2 = copy(tn)

julia> Tenet.@unsafe_region tn begin ### No errors thrown!
           tensor = Tensor(ones(3, 2), [:c, :d])
           push!(tn, tensor)
           push!(tn2, tensor)  # tn2 is not specified in @unsafe_region
           @test length(tensors(tn)) == 3
           pop!(tn, tensor)
       end
3×2 Tensor{Float64, 2, Matrix{Float64}}:
 1.0  1.0
 1.0  1.0
 1.0  1.0
```
In this code the `@unsafe_region` macro is applied only to `tn`. Inside the block, we also manipulate `tn2`, which is **not** specified in the macro. We expect `push!(tn2, tensor)` to throw a `DimensionMismatch` error due to inconsistent indices. However, no error is thrown, indicating that `@unsafe_region` is affecting `tn2` as well.

**Possible Solutions:**

1. Restrict the scope to specified `TensorNetwork`s:
  - Modify the `@unsafe_region` macro to ensure that it only affects the tensor networks explicitly listed.
2.  Change macro to reflect global scope:
  - If the intended behavior is to apply to all tensor networks within the block, we can consider changing the macro to not require specifying any tensor networks.
  
  For example:
```julia
@unsafe_region begin # no tn passed as argument here
    # Code here
end
```
This would make it clear to users that all tensor networks are affected within the block.

jofrevalles:
### Summary
This PR resolves #230 by refactoring the `@unsafe_region` macro. In this PR, we added a new field on the `TensorNetwork` struct `check_index_sizes::Ref{Bool}`. When `true`, there will be no inconsistent-dimension checks on the `push!` functions.

Now, `@unsafe_region` just modifies this value in the `field` of the specific `TensorNetwork` passed in the arguments, and keeps track of all the possible copies created inside the macro block.

### Example
Let's see how we fixed the example we show in #230:
```julia
julia> using Tenet

julia> tn = TensorNetwork([
           Tensor(ones(2, 2), [:a, :b]),
           Tensor(ones(2, 2), [:b, :c])
       ])

julia> tn2 = copy(tn)

julia> Tenet.@unsafe_region tn begin # Now it errors properly!
           tensor = Tensor(ones(3, 2), [:c, :d])
           push!(tn, tensor)
           push!(tn2, tensor)  # tn2 is not specified in @unsafe_region
           @test length(tensors(tn)) == 3
           pop!(tn, tensor)
       end
ERROR: DimensionMismatch: size(tensor,c)=3 but should be equal to size(tn,c)=2
Stacktrace:
 [1] push!(tn::TensorNetwork, tensor::Tensor{Float64, 2, Matrix{Float64}})
   @ Tenet ~/git/Tenet.jl/src/TensorNetwork.jl:381
 [2] macro expansion
   @ ./REPL[14]:4 [inlined]
 [3] top-level scope
```

But if instead `tn2` is created inside the macro block:
```julia
julia> tn = TensorNetwork([
                  Tensor(ones(2, 2), [:a, :b]),
                  Tensor(ones(2, 2), [:b, :c])
              ])

julia> Tenet.@unsafe_region tn begin # Here still errors since at the end `tn2` is inconsistent
                  tensor = Tensor(ones(3, 2), [:c, :d])
                  push!(tn, tensor)
                  tn2 = copy(tn)
                  push!(tn2, tensor)  # tn2 is not specified in @unsafe_region
                  @test length(tensors(tn)) == 3
                  pop!(tn, tensor)
       end
ERROR: DimensionMismatch: Inconsistent size of indices
Stacktrace:
 [1] top-level scope
   @ ~/git/Tenet.jl/src/TensorNetwork.jl:351

julia> Tenet.@unsafe_region tn begin ### No errors thrown!
                  tensor = Tensor(ones(3, 2), [:c, :d])
                  push!(tn, tensor)
                  tn2 = copy(tn)
                  push!(tn2, tensor)  # tn2 is not specified in @unsafe_region
                  @test length(tensors(tn)) == 3
                  pop!(tn, tensor)
                  pop!(tn2, tensor)
       end
3×2 Tensor{Float64, 2, Matrix{Float64}}:
 1.0  1.0
 1.0  1.0
 1.0  1.0
```
So we get the expected correct behavior.

### DIFF
--- a/src/TensorNetwork.jl
+++ b/src/TensorNetwork.jl
@@ -41,9 +41,10 @@ struct TensorNetwork <: AbstractTensorNetwork
     tensormap::IdDict{Tensor,Vector{Symbol}}
 
     sorted_tensors::CachedField{Vector{Tensor}}
-    check_index_sizes::Ref{Bool}
+    unsafe::Ref{Union{Nothing,UnsafeContext}}
 
-    function TensorNetwork(tensors; check_index_sizes::Union{Nothing,Bool}=nothing)
+    # TODO: Find a way to remove the `unsafe` keyword argument from the constructor
+    function TensorNetwork(tensors; unsafe::Union{Nothing,UnsafeContext}=nothing)
         tensormap = IdDict{Tensor,Vector{Symbol}}(tensor => inds(tensor) for tensor in tensors)
 
         indexmap = reduce(tensors; init=Dict{Symbol,Vector{Tensor}}()) do dict, tensor
@@ -54,25 +55,25 @@ struct TensorNetwork <: AbstractTensorNetwork
             dict
         end
 
-        if check_index_sizes === nothing || check_index_sizes == true
+        if isnothing(unsafe) # If there is not an active UnsafeContext, check for index sizes
             for ind in keys(indexmap) # Check for inconsistent index dimensions
                 dims = map(tensor -> size(tensor, ind), indexmap[ind])
                 length(unique(dims)) == 1 ||
                     throw(DimensionMismatch("Index $(ind) has inconsistent dimension: $(dims)"))
             end
 
-            return new(indexmap, tensormap, CachedField{Vector{Tensor}}(), Ref(true))
+            return new(indexmap, tensormap, CachedField{Vector{Tensor}}(), Ref{Union{Nothing,UnsafeContext}}(nothing))
         end
 
-        return new(indexmap, tensormap, CachedField{Vector{Tensor}}(), Ref(false))
+        return new(indexmap, tensormap, CachedField{Vector{Tensor}}(), Ref{Union{Nothing,UnsafeContext}}(unsafe))
     end
 end
 
 TensorNetwork() = TensorNetwork(Tensor[])
 TensorNetwork(tn::TensorNetwork) = tn
 
-function set_unsafe_context!(tn::TensorNetwork, uc::Bool)
-    tn.check_index_sizes[] = uc
+function set_unsafe_context!(tn::TensorNetwork, uc::Union{Nothing,UnsafeContext})
+    tn.unsafe[] = uc
     return tn
 end
 
@@ -82,12 +83,11 @@ end
 Return a shallow copy of a [`TensorNetwork`](@ref).
 """
 function Base.copy(tn::TensorNetwork)
-    new_tn = TensorNetwork(tensors(tn); check_index_sizes=tn.check_index_sizes[])
+    new_tn = TensorNetwork(tensors(tn); unsafe=tn.unsafe[])
 
-    # Check if there's an active UnsafeContext with tn in it
-    uc = current_unsafe_context()
-    if uc !== nothing && tn ∈ uc
-        push!(uc.refs, WeakRef(new_tn)) # Register the new copy in the UnsafeContext
+    # Check if there's an active UnsafeContext in the current tn
+    if !isnothing(tn.unsafe[])
+        push!(tn.unsafe[].refs, WeakRef(new_tn)) # Register the new copy in the UnsafeContext
     end
 
     return new_tn
@@ -281,18 +281,6 @@ function __check_index_sizes(tn)
     return true
 end
 
-struct UnsafeContext
-    refs::Vector{WeakRef}  # List of weak references
-
-    UnsafeContext() = new(Vector{WeakRef}())
-end
-
-Base.values(uc::UnsafeContext) = map(x -> x.value, uc.refs)
-
-# Global stack to manage nested unsafe contexts
-const _unsafe_context_stack = Ref{Vector{UnsafeContext}}(Vector{UnsafeContext}())
-
-Base.in(tn::TensorNetwork, tnstack::Vector{TensorNetwork}) = any(tn_ -> tn == tn_, tnstack)
 Base.in(tn::TensorNetwork, ucstack::Vector{UnsafeContext}) = any(uc -> tn ∈ values(uc), ucstack)
 Base.in(tn::TensorNetwork, uc::UnsafeContext) = tn ∈ values(uc)
 
@@ -311,16 +299,12 @@ macro unsafe_region(tn_sym, block)
         quote
             local old = copy($tn_sym)
 
-            # Create a new UnsafeContext and push it onto the stack
+            # Create a new UnsafeContext and set it to the current tn
             local _uc = Tenet.UnsafeContext()
-            push!(Tenet._unsafe_context_stack[], _uc)
-
-            # Set check_index_sizes to false for the passed tensor network
-            Tenet.set_unsafe_context!($tn_sym, false)
-            # $tn_sym.check_index_sizes[] = false
+            Tenet.set_unsafe_context!($tn_sym, _uc)
 
             # Register the tensor network in the context
-            push!(_uc.refs, WeakRef($tn_sym))
+            push!($tn_sym.unsafe[].refs, WeakRef($tn_sym))
 
             e = nothing
             try
@@ -331,19 +315,19 @@ macro unsafe_region(tn_sym, block)
             finally
                 if e === nothing
                     # Perform checks of registered tensor networks
-                    for ref in _uc.refs
+                    for ref in $tn_sym.unsafe[].refs
                         tn = ref.value
-                        if tn !== nothing
+                        if tn !== nothing && tn ∈ values($tn_sym.unsafe[])
                             if !Tenet.__check_index_sizes(tn)
                                 $(tn_sym) = old
-                                pop!(Tenet._unsafe_context_stack[])
+
+                                # Set `unsafe` field to `nothing`
+                                Tenet.set_unsafe_context!($tn_sym, nothing)
+
                                 throw(DimensionMismatch("Inconsistent size of indices"))
                             end
                         end
                     end
-
-                    # Pop the UnsafeContext from the stack
-                    pop!(Tenet._unsafe_context_stack[])
                 end
             end
         end,
@@ -361,9 +345,8 @@ function Base.push!(tn::AbstractTensorNetwork, tensor::Tensor)
     tn = TensorNetwork(tn)
     tensor ∈ keys(tn.tensormap) && return tn
 
-    # Check if there's an active UnsafeContext with tn in it
-    uc = current_unsafe_context()
-    if uc === nothing || tn ∉ uc  # Only check index sizes if we are not in an unsafe region
+    # Check if there's an active UnsafeContext in the tn
+    if isnothing(tn.unsafe[]) # Only check index sizes if we are not in an unsafe region
         for i in Iterators.filter(i -> size(tn, i) != size(tensor, i), inds(tensor) ∩ inds(tn))
             throw(
                 DimensionMismatch("size(tensor,$i)=$(size(tensor,i)) but should be equal to size(tn,$i)=$(size(tn,i))")
--- a/src/TensorNetwork.jl
+++ b/src/TensorNetwork.jl
@@ -41,10 +41,10 @@ struct TensorNetwork <: AbstractTensorNetwork
     tensormap::IdDict{Tensor,Vector{Symbol}}
 
     sorted_tensors::CachedField{Vector{Tensor}}
-    unsafe::Ref{Union{Nothing,UnsafeContext}}
+    unsafe::Ref{Union{Nothing,UnsafeScope}}
 
     # TODO: Find a way to remove the `unsafe` keyword argument from the constructor
-    function TensorNetwork(tensors; unsafe::Union{Nothing,UnsafeContext}=nothing)
+    function TensorNetwork(tensors; unsafe::Union{Nothing,UnsafeScope}=nothing)
         tensormap = IdDict{Tensor,Vector{Symbol}}(tensor => inds(tensor) for tensor in tensors)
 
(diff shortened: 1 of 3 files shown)


src/TensorNetwork.jl
```diff
@@ -59,7 +60,28 @@ struct TensorNetwork <: AbstractTensorNetwork
...
+            end
+            dict
+        end
+
+        if check_index_sizes
+            for ind in keys(indexmap)
+                dims = map(tensor -> size(tensor, ind), indexmap[ind])
+                length(unique(dims)) == 1 || throw(DimensionMismatch("Index $(ind) has inconsistent dimension: $(dims)"))
+            end
+        end
```
jofrevalles, line 82: I will group this so we don't have duplicated code. To do.

src/TensorNetwork.jl
```diff
@@ -261,20 +294,68 @@ function __check_index_sizes(tn)
...
+struct UnsafeContext
+    refs::Vector{WeakRef}  # List of weak references
+
+    UnsafeContext() = new(Vector{WeakRef}())
+end
+
+Base.values(uc::UnsafeContext) = map(x -> x.value, uc.refs)
+
+# Global stack to manage nested unsafe contexts
+const _unsafe_context_stack = Ref{Vector{UnsafeContext}}(Vector{UnsafeContext}())
```
mofeing, line 306: mmm you don't need this? the idea was that `@unsafe_region` first creates a `UnsafeContext` and passes that the `TensorNetwork`

src/TensorNetwork.jl
```diff
@@ -41,6 +41,7 @@ struct TensorNetwork <: AbstractTensorNetwork
     tensormap::IdDict{Tensor,Vector{Symbol}}
 
     sorted_tensors::CachedField{Vector{Tensor}}
+    check_index_sizes::Ref{Bool}
```
mofeing, line 44:
```suggestion
    check_index_sizes::Ref{UnsafeContext}
```

src/TensorNetwork.jl
```diff
@@ -59,7 +60,29 @@ struct TensorNetwork <: AbstractTensorNetwork
             length(unique(dims)) == 1 || throw(DimensionMismatch("Index $(ind) has inconsistent dimension: $(dims)"))
         end
 
-        return new(indexmap, tensormap, CachedField{Vector{Tensor}}())
+        return new(indexmap, tensormap, CachedField{Vector{Tensor}}(), Ref(true))
+    end
+
+    function TensorNetwork(tensors, check_index_sizes::Bool)
```
mofeing, line 66:
```suggestion
    function TensorNetwork(tensors, check_index_sizes)
```

src/TensorNetwork.jl
```diff
@@ -59,7 +60,29 @@ struct TensorNetwork <: AbstractTensorNetwork
...
+
+        if check_index_sizes
+            for ind in keys(indexmap)
+                dims = map(tensor -> size(tensor, ind), indexmap[ind])
+                length(unique(dims)) == 1 ||
+                    throw(DimensionMismatch("Index $(ind) has inconsistent dimension: $(dims)"))
+            end
+        end
+
+        return new(indexmap, tensormap, CachedField{Vector{Tensor}}(), Ref(check_index_sizes))
```
mofeing, line 85: i prefer another a function like `set_unsafe_context!(tn::TensorNetwork, ctx)` that sets that field, rather than another constructor

src/TensorNetwork.jl
```diff
@@ -71,7 +94,17 @@ TensorNetwork(tn::TensorNetwork) = tn
 
 Return a shallow copy of a [`TensorNetwork`](@ref).
 """
-Base.copy(tn::TensorNetwork) = TensorNetwork(tensors(tn))
+function Base.copy(tn::TensorNetwork)
+    new_tn = TensorNetwork(tensors(tn), tn.check_index_sizes[])
```
mofeing, line 98: with the comment above, you should rather do:
```suggestion
    new_tn = TensorNetwork(tensors(tn))
    set_unsafe_context!(new_tn, tn.check_index_sizes[])
```

src/TensorNetwork.jl
```diff
@@ -261,20 +294,68 @@ function __check_index_sizes(tn)
...
+
+    UnsafeContext() = new(Vector{WeakRef}())
+end
+
+Base.values(uc::UnsafeContext) = map(x -> x.value, uc.refs)
+
+# Global stack to manage nested unsafe contexts
+const _unsafe_context_stack = Ref{Vector{UnsafeContext}}(Vector{UnsafeContext}())
+
+Base.in(tn::TensorNetwork, tnstack::Vector{TensorNetwork}) = any(tn_ -> tn === tn_, tnstack)
```
mofeing, line 308: Careful, this method is overloading too much stuff. Semantically `in` and `===` are different: `in` uses `==`.

src/TensorNetwork.jl
jofrevalles: You need this to check if there is an active `UnsafeContext` when you call `copy` and related functions.

src/TensorNetwork.jl
jofrevalles: You can not do that since there is a index check inside the constructor.

src/TensorNetwork.jl
jofrevalles: You need this to check if there is an active `UnsafeContext` when you call copy and related functions.

src/TensorNetwork.jl
```diff
@@ -53,25 +54,44 @@ struct TensorNetwork <: AbstractTensorNetwork
...
     copy(tn::TensorNetwork)
 
 Return a shallow copy of a [`TensorNetwork`](@ref).
 """
-Base.copy(tn::TensorNetwork) = TensorNetwork(tensors(tn))
+function Base.copy(tn::TensorNetwork)
+    new_tn = TensorNetwork(tensors(tn); check_index_sizes=tn.check_index_sizes[])
+
+    # Check if there's an active UnsafeContext with tn in it
+    uc = current_unsafe_context()
```
mofeing, line 88: This is what I want to talk to you about tomorrow. We just need to make a decision between 2 options.

src/TensorNetwork.jl
```diff
@@ -261,20 +281,69 @@ function __check_index_sizes(tn)
...
+    UnsafeContext() = new(Vector{WeakRef}())
+end
+
+Base.values(uc::UnsafeContext) = map(x -> x.value, uc.refs)
 
-macro unsafe_region(tn, block)
+# Global stack to manage nested unsafe contexts
+const _unsafe_context_stack = Ref{Vector{UnsafeContext}}(Vector{UnsafeContext}())
+
+Base.in(tn::TensorNetwork, tnstack::Vector{TensorNetwork}) = any(tn_ -> tn == tn_, tnstack)
```
mofeing, line 295: mmmm do we need this? this should be automatic

mofeing (commented):
In general, I really like it. It's quite elegant the way it solves.

Just want to talk to you about the global variable where it stores the current unsafe context.

jofrevalles:
@mofeing global `const` is gone and everything looks nice now :)

src/TensorNetwork.jl
```diff
@@ -261,20 +281,52 @@ function __check_index_sizes(tn)
...
-macro unsafe_region(tn, block)
+# Function to get the current UnsafeContext
+function current_unsafe_context()
+    if isempty(Tenet._unsafe_context_stack[])
+        return nothing
+    else
+        return Tenet._unsafe_context_stack[][end]
+    end
+end
+
```
mofeing, line 294: This function is no longer need right?
```suggestion
```

src/TensorNetwork.jl
```diff
@@ -261,20 +281,52 @@ function __check_index_sizes(tn)
...
+# Define the @unsafe_region macro
```
mofeing, line 295: this comment is a lil bit auto-descriptive? maybe we can just remove it or add a docstring

src/TensorNetwork.jl
```diff
@@ -292,8 +344,8 @@ function Base.push!(tn::AbstractTensorNetwork, tensor::Tensor)
     tn = TensorNetwork(tn)
     tensor ∈ keys(tn.tensormap) && return tn
 
-    # Only check index sizes if we are not in an unsafe region
-    if !is_unsafe_region[]
+    # Check if there's an active UnsafeContext in the tn
+    if isnothing(tn.unsafe[]) # Only check index sizes if we are not in an unsafe region
```
mofeing, line 348:
```suggestion
    # Check index sizes if there isn't an active `UnsafeContext` in the Tensor Network
    if isnothing(tn.unsafe[])
```

src/TensorNetwork.jl
```diff
@@ -53,25 +55,43 @@ struct TensorNetwork <: AbstractTensorNetwork
             dict
         end
 
-        # Check for inconsistent index dimensions
-        for ind in keys(indexmap)
-            dims = map(tensor -> size(tensor, ind), indexmap[ind])
-            length(unique(dims)) == 1 || throw(DimensionMismatch("Index $(ind) has inconsistent dimension: $(dims)"))
+        if isnothing(unsafe) # If there is not an active UnsafeContext, check for index sizes
+            for ind in keys(indexmap) # Check for inconsistent index dimensions
```
mofeing, line 59:
```suggestion
        # check index size consistency if not inside an `UnsafeContext`
        if isnothing(unsafe)
            for ind in keys(indexmap)
```

src/TensorNetwork.jl
```diff
@@ -53,25 +55,43 @@ struct TensorNetwork <: AbstractTensorNetwork
...
+                dims = map(tensor -> size(tensor, ind), indexmap[ind])
+                length(unique(dims)) == 1 ||
+                    throw(DimensionMismatch("Index $(ind) has inconsistent dimension: $(dims)"))
+            end
+
+            return new(indexmap, tensormap, CachedField{Vector{Tensor}}(), Ref{Union{Nothing,UnsafeContext}}(nothing))
```
mofeing, line 65: You can rely on the constructor called later because `unsafe` is `nothing`

```suggestion
```

src/TensorNetwork.jl
```diff
@@ -53,25 +55,43 @@ struct TensorNetwork <: AbstractTensorNetwork
...
 
 Return a shallow copy of a [`TensorNetwork`](@ref).
 """
-Base.copy(tn::TensorNetwork) = TensorNetwork(tensors(tn))
+function Base.copy(tn::TensorNetwork)
+    new_tn = TensorNetwork(tensors(tn); unsafe=tn.unsafe[])
+
+    # Check if there's an active UnsafeContext in the current tn
+    if !isnothing(tn.unsafe[])
+        push!(tn.unsafe[].refs, WeakRef(new_tn)) # Register the new copy in the UnsafeContext
```
mofeing, line 90:
```suggestion
    # register a new `TensorNetwork` to be signaled when the `UnsafeContext` goes out of scope
    if !isnothing(tn.unsafe[])
        push!(tn.unsafe[].refs, WeakRef(new_tn))
```

src/TensorNetwork.jl
```diff
@@ -53,25 +55,43 @@ struct TensorNetwork <: AbstractTensorNetwork
...
     end
 end
 
 TensorNetwork() = TensorNetwork(Tensor[])
 TensorNetwork(tn::TensorNetwork) = tn
 
+function set_unsafe_context!(tn::TensorNetwork, uc::Union{Nothing,UnsafeContext})
+    tn.unsafe[] = uc
+    return tn
+end
```
mofeing, line 78: Can you generalize this to `AbstractTensorNetwork`?
```suggestion
function set_unsafe_context!(tn::AbstractTensorNetwork, uc::Union{Nothing,UnsafeContext})
    TensorNetwork(tn).unsafe[] = uc
    return tn
end
```

mofeing (changes requested):
Perfect! Aside from the small requests, would you mind doing 2 more extra things?

1. Rename it to `UnsafeScope`: The word context has too many meanings and `@unsafe_region` is actually a scope. It also took from inspiration from the ScopedValues.jl package.
2. Implement a `get_unsafe_context` method to retrieve it from any `AbstractTensorNetwork`: This way we wouldn't need to call `TensorNetwork(tn)` on the beginning of `@unsafe_region`. You will need to change any reference to `tn.unsafe` to `get_unsafe_context(tn)`.

```julia
get_unsafe_context(tn::AbstractTensorNetwork) = TensorNetwork(tn).unsafe
```

src/TensorNetwork.jl
jofrevalles: Maybe its better to remove it. I prefer not to have a `docstring` here since this is an internal operation, for now.

jofrevalles:
@mofeing I added some of your suggestions and changed the name to `UnsafeScope`, I think this is ready for merge.

mofeing (approved):

PR #248, opened 2024-11-13
Add `Form`-specific constructors for `MPS`

jofrevalles:
This PR adds `Form`-specific constructors for the `MPS` struct, which allows us to create Matrix Product States with an specific canonical `Form` directly on the constructor. I believe this is the proper way to go, and we added a `check_canonical_form` `kwarg` that, if `true` (default), it will check if the `MPS` fulfills the canonical form.

### TO DO:

- [x] ~~Regroup repeated code into an internal function. Maybe it can be named `__tn_mps_from_arrays`? @mofeing I can't find a good name.~~ Group code by using `NonCanonical` constructor.

### DIFF
--- a/src/MPS.jl
+++ b/src/MPS.jl
@@ -38,6 +38,9 @@ Base.zero(x::T) where {T<:Union{MPS,MPO}} = T(zero(Ansatz(x)), form(x))
 defaultorder(::Type{<:AbstractMPS}) = (:o, :l, :r)
 defaultorder(::Type{<:AbstractMPO}) = (:o, :i, :l, :r)
 
+MPS(arrays::Vector{<:AbstractArray}; order=defaultorder(MPS), form::Form=NonCanonical(), check_canonical_form = true) = MPS(form, arrays; order=order, check_canonical_form=check_canonical_form)
+MPS(arrays::Vector{<:AbstractArray}, λ::Vector{<:AbstractArray}; order=defaultorder(MPS), form::Form=Canonical(), check_canonical_form = true) = MPS(form, arrays, λ; order=order, check_canonical_form=check_canonical_form)
+
 """
     MPS(arrays::Vector{<:AbstractArray}; order=defaultorder(MPS))
 
@@ -47,7 +50,7 @@ Create a [`NonCanonical`](@ref) [`MPS`](@ref) from a vector of arrays.
 
   - `order` The order of the indices in the arrays. Defaults to `(:o, :l, :r)`.
 """
-function MPS(arrays::Vector{<:AbstractArray}; order=defaultorder(MPS))
+function MPS(::NonCanonical, arrays; order=defaultorder(MPS), check_canonical_form = true)
     @assert ndims(arrays[1]) == 2 "First array must have 2 dimensions"
     @assert all(==(3) ∘ ndims, arrays[2:(end - 1)]) "All arrays must have 3 dimensions"
     @assert ndims(arrays[end]) == 2 "Last array must have 2 dimensions"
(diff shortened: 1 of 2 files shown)


src/MPS.jl
```diff
@@ -38,6 +38,21 @@ Base.zero(x::T) where {T<:Union{MPS,MPO}} = T(zero(Ansatz(x)), form(x))
...
+function MPS(
+    arrays::Vector{<:AbstractArray},
+    λ::Vector{<:AbstractArray};
+    order=defaultorder(MPS),
+    form::Form=Canonical(),
+    check_canonical_form=true,
+)
+    return MPS(form, arrays, λ; order=order, check_canonical_form=check_canonical_form)
+end
+
```
mofeing, line 55: 1. I would just call it `check`. `check_canonical_form` is too long and for the size checks we already use the `@unsafe_region`.
2. If the variable you're passing to the kwarg is named the same as the kwarg, you don't to write the `=`. It's cleaner this way.
3. For default kwargs that are just passed by, use `kwargs...`
4. If they are not really needed, don't put the types in the args.
  - Types in function signatures should only (mostly) be used for dispatch

```suggestion
MPS(arrays; form::Form=NonCanonical(), kwargs...) = MPS(form, arrays; kwargs...)
function MPS(arrays, λ; form::Form=Canonical(), kwargs...)
    return MPS(form, arrays, λ; kwargs...)
end

```

mofeing (changes requested):
Great! Thanks! In general everything is great. Just some suggestions for the generic `MPS` constructors.

> Regroup repeated code into an internal function. Maybe it can be named `__tn_mps_from_arrays`?

You no longer need to prefix internal functions with `__` since we'll soon start annotating public API with the new `public` keyword (well, with the `@public` macro from Compat.jl).

Also, what I would recommend you is not to refactor it to an internal function, but to leave that code in the `NonCanonical` constructor. Then `Canonical` and `MixedCanonical` would call the `NonCanonical` constructor and modify it.

- `MixedCanonical` calls the `NonCanonical` constructor and then modifies the `form` field.
- `Canonical` calls the `NonCanonical` constructor, tensorizes the $\Alpha$ by requesting bond index name with `inds(tn; bond=...)`, push them to the TN and finally modify the `form` field.

An example of this pattern can be found in `simple_update!` for `MixedCanonical`: it first calls `mixed_canonize!` to move the orthogonality center to the target sites and then calls `NonCanonical` version of `simple_update!`.

src/MPS.jl
jofrevalles: I need to add the types in the args because we need to distinguish from the `MPS(ansatz::Ansatz, form::Form)` constructor. Other than that, I agree with your suggestions!

jofrevalles:
Okay, everything should be ready now @mofeing. I separated the checks of the form in two `check_form` functions that are dispatched over the `Form`. I think this is a good idea since we would want to use these functions somewhere outside the constructor. In this case, maybe it's even better that they do not `throw` an error but just return `false`? I don't know, for me this is still fine.

src/MPS.jl
```diff
@@ -204,7 +267,7 @@ Base.adjoint(tn::T) where {T<:AbstractMPO} = T(adjoint(Ansatz(tn)), form(tn))
 """
     Base.rand(rng::Random.AbstractRNG, ::Type{MPS}; n, maxdim, eltype=Float64, physdim=2)
 
-Create a random [`MPS`](@ref) Tensor Network.
+Create a random [`MPS`](@ref) Tensor Network where all tensors are left-canonical (MixedCanonical(Site(0))).
```
mofeing, line 270: Should be right-canonical right?

```suggestion
Create a random [`MPS`](@ref) Tensor Network where all tensors are right-canonical (orthogonality center at site 1).
```

src/MPS.jl
```diff
@@ -238,7 +301,8 @@ function Base.rand(rng::Random.AbstractRNG, ::Type{MPS}; n, maxdim, eltype=Float
     arrays[1] = reshape(arrays[1], p, p)
     arrays[n] = reshape(arrays[n], p, p)
 
-    return MPS(arrays; order=(:l, :o, :r))
+    return MPS(arrays; order=(:l, :o, :r), form=MixedCanonical(Site(0)))
+    # return MPS(arrays; order=(:l, :o, :r))
```
mofeing, line 305:
```suggestion
    return MPS(arrays; order=(:l, :o, :r), form=MixedCanonical(site"1"))
```

src/MPS.jl
```diff
@@ -92,6 +97,64 @@ function MPS(arrays::Vector{<:AbstractArray}; order=defaultorder(MPS))
...
+function check_form(::Canonical, mps::AbstractMPO)
+    for i in 1:nsites(mps)
+        if i > 1
+            isisometry(contract(mps; between=(Site(i - 1), Site(i)), direction=:right), Site(i); dir=:right) ||
+                throw(ArgumentError("Can not form a left-canonical tensor in Site($i) from Γ and λ contraction."))
+        end
+
+        if i < nsites(mps)
+            isisometry(contract(mps; between=(Site(i), Site(i + 1)), direction=:left), Site(i); dir=:left) ||
+                throw(ArgumentError("Can not form a right-canonical tensor in Site($i) from Γ and λ contraction."))
```
mofeing, line 151: I think it's clearer if you fuse these 2 conditions:
```suggestion
        if i < nsites(mps) && isisometry(contract(mps; between=(Site(i), Site(i + 1)), direction=:left), Site(i); dir=:left)
            throw(ArgumentError("Can not form a right-canonical tensor in site $i from Γ and λ contraction."))
```

src/MPS.jl
```diff
@@ -92,6 +97,64 @@ function MPS(arrays::Vector{<:AbstractArray}; order=defaultorder(MPS))
...
+    end
+
+    return true
+end
+
+function check_form(::Canonical, mps::AbstractMPO)
+    for i in 1:nsites(mps)
+        if i > 1
+            isisometry(contract(mps; between=(Site(i - 1), Site(i)), direction=:right), Site(i); dir=:right) ||
+                throw(ArgumentError("Can not form a left-canonical tensor in Site($i) from Γ and λ contraction."))
```
mofeing, line 146: Same as below:
```suggestion
        if i > 1 && isisometry(contract(mps; between=(Site(i - 1), Site(i)), direction=:right), Site(i); dir=:right)
            throw(ArgumentError("Can not form a left-canonical tensor in Site($i) from Γ and λ contraction."))
```

mofeing (commented):
Great! Just some suggestions and we're done.

mofeing (approved):

mofeing:
Thanks!

PR #253, opened 2024-11-19
Fix `Quantum` function for `Quac.Gate` type

jofrevalles:
### Summary
This PR fixes a problem with the `Quantum` function for `Quac.Gate` argument. Before this PR, this function had legacy code from our previous `Dense` structure, which was replaced in PR #223. Specifically, we still had the `sites` kwarg, which is now replaced by the second argument of `Quantum`.

Additionally, in this PR we also add a `@testset` `"Gate"` in the `Quac` integration tests to cover for this issue.


mofeing (approved):
Great! Thanks!

PR #254, opened 2024-11-19
Fix `truncate!` function

jofrevalles:
### Summary
This PR resolves a bug found in the `truncate!` function that erroed when `maxdim` kwarg was larger than the `size` of the bond that was about to truncate. In a similar way, it also erroed when the `threshold` was lower than the smaller element in the `spectrum` that was used to truncate.

Moreover, we fixed the `mixed_canonize!` function, since previously did not work as the `truncate!` function expected for `MPS` with `MixedCanonical` forms. We also fixed an unnecessary canonization step in the `truncate!` for `Canonical` forms.

### Example
Without this PR:
```julia
julia> using Tenet; using Test; using LinearAlgebra

julia> gate = Quantum(reshape(LinearAlgebra.I(4), 2, 2, 2, 2), [site"2", site"3", site"2'", site"3'"])
Quantum (inputs=2, outputs=2)

julia> ψ = MPS([rand(2, 2), rand(2, 2, 2), rand(2, 2, 2), rand(2, 2)])
MPS (inputs=0, outputs=4)

julia> ψ = deepcopy(ψ)
MPS (inputs=0, outputs=4)

julia> canonize!(ψ)
MPS (inputs=0, outputs=4)

julia> evolved = evolve!(deepcopy(ψ), gate; maxdim=10)
spectrum = [7.129434737170108, 1.2106678577680343, 3.16516617082361e-16, 9.749816418484561e-18]
ERROR: BoundsError: attempt to access 2×2×4 Array{Float64, 3} at index [1:2, 1:2, 1:10]
Stacktrace: ....

julia> evolved = evolve!(deepcopy(ψ), gate; threshold=1e-100)
ERROR: MethodError: no method matching -(::Nothing, ::Int64)
The function `-` exists, but no method is defined for this combination of argument types.

Closest candidates are: ...
```

Everything now works, and we extended the tests to cover for this changes.

### DIFF
--- a/src/Ansatz.jl
+++ b/src/Ansatz.jl
@@ -255,8 +255,8 @@ Truncate the dimension of the virtual `bond`` of an [`Ansatz`](@ref) Tensor Netw
 
   - Either `threshold` or `maxdim` must be provided. If both are provided, `maxdim` is used.
 """
-function truncate!(tn::AbstractAnsatz, bond; threshold=nothing, maxdim=nothing)
-    return truncate!(form(tn), tn, bond; threshold, maxdim)
+function truncate!(tn::AbstractAnsatz, bond; threshold=nothing, maxdim=nothing, kwargs...)
+    return truncate!(form(tn), tn, bond; threshold, maxdim, kwargs...)
 end
 
 """
@@ -315,8 +315,18 @@ function truncate!(::MixedCanonical, tn::AbstractAnsatz, bond; threshold, maxdim
     return truncate!(NonCanonical(), tn, bond; threshold, maxdim, compute_local_svd=true)
 end
 
-function truncate!(::Canonical, tn::AbstractAnsatz, bond; threshold, maxdim)
-    return truncate!(NonCanonical(), tn, bond; threshold, maxdim, compute_local_svd=false)
+"""
+    truncate!(::Canonical, tn::AbstractAnsatz, bond; threshold, maxdim, recanonize=true)
+
+Truncate the dimension of the virtual `bond` of a [`Canonical`](@ref) Tensor Network by keeping the `maxdim` largest
+**Schmidt coefficients** or those larger than `threshold`, and then recanonizes the Tensor Network if `recanonize` is `true`.
+"""
+function truncate!(::Canonical, tn::AbstractAnsatz, bond; threshold, maxdim, recanonize=true)
+    truncate!(NonCanonical(), tn, bond; threshold, maxdim, compute_local_svd=false)
+
+    recanonize && canonize!(tn)
+
+    return tn
 end
 
 overlap(a::AbstractAnsatz, b::AbstractAnsatz) = contract(merge(a, copy(b)'))
--- a/src/Ansatz.jl
+++ b/src/Ansatz.jl
@@ -49,7 +49,7 @@ struct NonCanonical <: Form end
     left of the orthogonality center are left-canonical and the tensors to the right are right-canonical.
 """
 struct MixedCanonical <: Form
-    orthog_center::Union{Site,Vector{Site}}
+    orthog_center::Union{Site,Vector{<:Site}}
 end
 
 """
--- a/src/Ansatz.jl
+++ b/src/Ansatz.jl
@@ -312,7 +312,7 @@ end
 function truncate!(::MixedCanonical, tn::AbstractAnsatz, bond; threshold, maxdim)
     # move orthogonality center to bond
     mixed_canonize!(tn, bond)
-    return truncate!(NonCanonical(), tn, bond; threshold, maxdim, compute_local_svd=false)
+    return truncate!(NonCanonical(), tn, bond; threshold, maxdim, compute_local_svd=true)
 end
 
 function truncate!(::Canonical, tn::AbstractAnsatz, bond; threshold, maxdim)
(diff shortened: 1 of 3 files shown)


src/Ansatz.jl
```diff
@@ -308,13 +312,11 @@ end
 function truncate!(::MixedCanonical, tn::AbstractAnsatz, bond; threshold, maxdim)
     # move orthogonality center to bond
     mixed_canonize!(tn, bond)
-    return truncate!(NonCanonical(), tn, bond; threshold, maxdim, compute_local_svd=false)
+    return truncate!(NonCanonical(), tn, bond; threshold, maxdim, compute_local_svd=true)
```
mofeing, line 315: mmm why is `compute_local_svd` required?

src/MPS.jl
```diff
@@ -504,18 +504,30 @@ end
...
+    if orthog_center isa Site
+        left = id(orthog_center) - 1
+        right = id(orthog_center) + 1
+    else
+        values = [id(site) for site in orthog_center]
+        orthog_center = Vector{Site}(orthog_center)
+
+        left, right = extrema(values) .+ (-1, 1)
+    end
+
```
mofeing, line 516: how about this? you don't need to collect `orthog_center` right?

```suggestion
    left, right = if orthog_center isa Site
        id(orthog_center) .+ (-1, 1)
    elseif orthog_center isa Vector{<:Site}
        extrema(id.(orthog_center)) .+ (-1, 1)
    else
        throw(ArgumentError("`orthog_center` must be a `Site` or a `Vector{Site}`")
    end

```

src/MPS.jl
```diff
@@ -504,18 +504,30 @@ end
...
-    for i in nsites(tn):-1:(id(orthog_center) + 1)
+    for i in nsites(tn):-1:right
         canonize_site!(tn, Site(i); direction=:left, method=:qr)
     end
 
     # center SVD sweep to get singular values
-    # canonize_site!(tn, orthog_center; direction=:left, method=:svd)
+    # for i in (left + 1):(right - 1)
+    #     canonize_site!(tn, Site(i); direction=:left, method=:svd)
+    # end
```
mofeing, line 530: with the orthogonality center not being a unique site, this makes little sense right?

what do you think? should we remove this comment and later move it to the `BondMixedCanonical`?

src/Ansatz.jl
```diff
@@ -308,13 +312,11 @@ end
...
 end
 
 function truncate!(::Canonical, tn::AbstractAnsatz, bond; threshold, maxdim)
-    truncate!(NonCanonical(), tn, bond; threshold, maxdim, compute_local_svd=false)
-    # requires a sweep to recanonize the TN
-    return canonize!(tn)
+    return truncate!(NonCanonical(), tn, bond; threshold, maxdim, compute_local_svd=false)
```
mofeing, line 319: mmm you're missing the contraction with lateral $\Lambda$ and the reconstruction no?

src/Ansatz.jl
jofrevalles: Yes, since your implementation of `MixedCanonical` does not have the singular values, you need to do the `svd` to be able to truncate

src/Ansatz.jl
jofrevalles: I don't see why this is necessary, when you have already the things in the `Canonical` form the singular values in the bonds can already be used to `truncate` the `MPS`, since you can view like if you are in the orthogonality center.

src/MPS.jl
jofrevalles: ahh I think I did that because if `orthog_center` in the argument is something like `[Site(2), Site(3)]`, this has type of `Vector{Site{1}}` which does not enter on the constructor function which is defined for `Union{Site, Vector{Site}}`, that's why I had that `Vector{Site}(orthog_center)` part.

src/MPS.jl
jofrevalles: yes, for me it is okay to remove the comment! I will do that tomorrow in a commit :)

src/Ansatz.jl
jofrevalles: right, @starsfordummies ?

starsfordummies (approved):
nice!

src/MPS.jl
```diff
@@ -126,14 +126,26 @@ function MPS(::Canonical, arrays, λ; order=defaultorder(MPS), check=true)
...
+
+Check if the tensors in the mps are in the proper [`Form`](@ref).
+"""
 check_form(mps::AbstractMPO) = check_form(form(mps), mps)
 
 function check_form(config::MixedCanonical, mps::AbstractMPO)
     orthog_center = config.orthog_center
+
+    left, right = if orthog_center isa Site
+        id(orthog_center) .+ (0, 0)
```
mofeing, line 140: mmm remove the add?
```suggestion
        id(orthog_center)
```

mofeing (approved):
Everything perfect! Just one suggestion to remove `.+ (0,0)` because it does nothing.

src/MPS.jl
jofrevalles: This is needed to unpack the same value to both `left` and `right` 🥶

mofeing: ah okay, can you put a comment then?

jofrevalles: oki

PR #255, opened 2024-11-20
Enhance `simple_update!` for `MPS` in the `Canonical` form

jofrevalles:
### Summary
This PR refactors the `simple_update!` function so now it properly uses the $\Lambda$ tensors of an `MPS` in the `Canonical` form to `evolve` the state without the need of recanonizing it.

Additionally, we also added the `kwarg` of `normalize` in the `canonize!` function.

### DIFF
--- a/src/Ansatz.jl
+++ b/src/Ansatz.jl
@@ -321,7 +321,7 @@ end
 Truncate the dimension of the virtual `bond` of a [`Canonical`](@ref) Tensor Network by keeping the `maxdim` largest
 **Schmidt coefficients** or those larger than `threshold`, and then recanonizes the Tensor Network if `recanonize` is `true`.
 """
-function truncate!(::Canonical, tn::AbstractAnsatz, bond; threshold, maxdim, recanonize=true)
+function truncate!(::Canonical, tn::AbstractAnsatz, bond; threshold, maxdim, recanonize=false)
     truncate!(NonCanonical(), tn, bond; threshold, maxdim, compute_local_svd=false)
 
     recanonize && canonize!(tn; normalize=true)
--- a/src/Ansatz.jl
+++ b/src/Ansatz.jl
@@ -274,8 +274,11 @@ Truncate the dimension of the virtual `bond` of a [`NonCanonical`](@ref) Tensor 
     - `threshold`: The threshold to truncate the bond dimension.
     - `maxdim`: The maximum bond dimension to keep.
     - `compute_local_svd`: Whether to compute the local SVD of the bond. If `true`, it will contract the bond and perform a SVD to get the local singular values. Defaults to `true`.
+    - `renormalize`: Whether to renormalize the state at the bond after truncation. Defaults to `false`.
 """
-function truncate!(::NonCanonical, tn::AbstractAnsatz, bond; threshold, maxdim, compute_local_svd=true)
+function truncate!(
+    ::NonCanonical, tn::AbstractAnsatz, bond; threshold, maxdim, compute_local_svd=true, renormalize=false
+)
     virtualind = inds(tn; bond)
 
     if compute_local_svd
@@ -305,14 +308,18 @@ function truncate!(::NonCanonical, tn::AbstractAnsatz, bond; threshold, maxdim, 
     end
 
     slice!(tn, virtualind, extent)
+    sliced_bond = tensors(tn; bond)
+
+    renormalize && replace!(tn, sliced_bond => sliced_bond ./ norm(tn))
 
     return tn
 end
 
-function truncate!(::MixedCanonical, tn::AbstractAnsatz, bond; threshold, maxdim)
+function truncate!(::MixedCanonical, tn::AbstractAnsatz, bond; threshold, maxdim, renormalize=false)
     # move orthogonality center to bond
     mixed_canonize!(tn, bond)
-    return truncate!(NonCanonical(), tn, bond; threshold, maxdim, compute_local_svd=true)
+
+    return truncate!(NonCanonical(), tn, bond; threshold, maxdim, compute_local_svd=true, renormalize=renormalize)
 end
 
 """
@@ -321,10 +328,10 @@ end
 Truncate the dimension of the virtual `bond` of a [`Canonical`](@ref) Tensor Network by keeping the `maxdim` largest
 **Schmidt coefficients** or those larger than `threshold`, and then recanonizes the Tensor Network if `recanonize` is `true`.
 """
-function truncate!(::Canonical, tn::AbstractAnsatz, bond; threshold, maxdim, recanonize=false)
-    truncate!(NonCanonical(), tn, bond; threshold, maxdim, compute_local_svd=false)
+function truncate!(::Canonical, tn::AbstractAnsatz, bond; threshold, maxdim, recanonize=false, renormalize=renormalize)
+    truncate!(NonCanonical(), tn, bond; threshold, maxdim, compute_local_svd=false, renormalize=renormalize)
 
-    recanonize && canonize!(tn; normalize=true)
+    recanonize && canonize!(tn)
 
     return tn
 end
--- a/src/Ansatz.jl
+++ b/src/Ansatz.jl
@@ -419,6 +419,9 @@ function simple_update_1site!(ψ::AbstractAnsatz, gate)
     return contract!(ψ, contracting_index)
 end
 
+simple_update_2site!(::MixedCanonical, ψ::AbstractAnsatz, gate; threshold=nothing, maxdim=nothing, renormalize=false) =
+    simple_update_2site!(NonCanonical(), ψ, gate; threshold, maxdim, renormalize)
+
 # TODO remove `renormalize` argument?
 function simple_update_2site!(
     ::NonCanonical, ψ::AbstractAnsatz, gate; threshold=nothing, maxdim=nothing, renormalize=false
@@ -456,7 +459,7 @@ function simple_update_2site!(
 
     # truncate virtual index
     if any(!isnothing, (threshold, maxdim))
-        truncate!(ψ, bond; threshold, maxdim)
+        truncate!(ψ, [bond...]; threshold, maxdim)
         renormalize && normalize!(ψ, bond[1])
     end
 
--- a/src/Ansatz.jl
+++ b/src/Ansatz.jl
@@ -387,11 +387,11 @@ function simple_update!(ψ::AbstractAnsatz, gate; threshold=nothing, maxdim=noth
 
     if nlanes(gate) == 1
         return simple_update_1site!(ψ, gate)
+    elseif nlanes(gate) == 2
+        return simple_update_2site!(form(ψ), ψ, gate; threshold, maxdim, kwargs...)
+    else
+        throw(ArgumentError("Only 1-site and 2-site gates are currently supported"))
     end
-
-    @assert has_edge(ψ, lanes(gate)...) "Gate must act on neighboring sites"
-
-    return simple_update!(form(ψ), ψ, gate; threshold, maxdim, kwargs...)
 end
 
 # TODO a lot of problems with merging... maybe we shouldn't merge manually
@@ -420,8 +420,9 @@ function simple_update_1site!(ψ::AbstractAnsatz, gate)
 end
 
 # TODO remove `renormalize` argument?
-function simple_update!(::NonCanonical, ψ::AbstractAnsatz, gate; threshold=nothing, maxdim=nothing, renormalize=false)
-    @assert nlanes(gate) == 2 "Only 2-site gates are supported currently"
+function simple_update_2site!(
+    ::NonCanonical, ψ::AbstractAnsatz, gate; threshold=nothing, maxdim=nothing, renormalize=false
+)
     @assert has_edge(ψ, lanes(gate)...) "Gate must act on neighboring sites"
 
     # shallow copy to avoid problems if errors in mid execution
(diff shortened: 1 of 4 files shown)


jofrevalles:
@mofeing @starsfordummies Maybe we should put the `renormalize` kwarg in all `truncate!` functions (for all the `Form`s), right? Since we will lose norm every time we truncate.

starsfordummies:
ok for me!

src/Ansatz.jl
```diff
@@ -321,10 +321,10 @@ end
 Truncate the dimension of the virtual `bond` of a [`Canonical`](@ref) Tensor Network by keeping the `maxdim` largest
 **Schmidt coefficients** or those larger than `threshold`, and then recanonizes the Tensor Network if `recanonize` is `true`.
 """
-function truncate!(::Canonical, tn::AbstractAnsatz, bond; threshold, maxdim, recanonize=true)
+function truncate!(::Canonical, tn::AbstractAnsatz, bond; threshold, maxdim, recanonize=false)
```
mofeing, line 324: I would put it `recanonize` to `true` by default because in the general case, truncating would make the canonical ill-formed (so not really a canonical form)

src/Ansatz.jl
```diff
@@ -321,10 +321,10 @@ end
...
     truncate!(NonCanonical(), tn, bond; threshold, maxdim, compute_local_svd=false)
 
-    recanonize && canonize!(tn)
+    recanonize && canonize!(tn; normalize=true)
```
mofeing, line 327: like you said in the comment, this should be controlled by a kwarg

src/Ansatz.jl
```diff
@@ -455,16 +462,38 @@ function simple_update!(::NonCanonical, ψ::AbstractAnsatz, gate; threshold=noth
 
     # truncate virtual index
     if any(!isnothing, (threshold, maxdim))
-        truncate!(ψ, bond; threshold, maxdim)
+        truncate!(ψ, [bond...]; threshold, maxdim)
```
mofeing, line 465:
```suggestion
        truncate!(ψ, collect(bond); threshold, maxdim)
```

src/MPS.jl
```diff
@@ -495,6 +495,7 @@ function canonize!(ψ::AbstractMPO)
 
         # extract the singular values and contract them with the next tensor
         Λᵢ = pop!(ψ, tensors(ψ; between=(Site(i), Site(i + 1))))
+        normalize && (Λᵢ ./= norm(Λᵢ))
```
mofeing, line 498: I'm not sure why you need to normalize on canonization? Like it's completely doable, but I would prefer them to be separate because semantically are different things and there is no performance overhead when doing this:

```julia
canonize!(ψ)
normalize!(ψ)
```

src/MPS.jl
```diff
@@ -556,4 +557,4 @@ function LinearAlgebra.normalize!(config::MixedCanonical, ψ::AbstractMPO; at=co
     return ψ
 end
 
-# TODO function LinearAlgebra.normalize!(::Canonical, ψ::AbstractMPO) end
+LinearAlgebra.normalize!(::Canonical, ψ::AbstractMPO) = canonize!(ψ; normalize=true)
```
mofeing, line 560: mmm why do you need to canonize on normalization?

src/MPS.jl
jofrevalles: Because it's the way to normalize a state when you have an `mps` in the `Canonical` `Form`. This makes much more sense than what we have now, which involves converting the `Canonical` into `MixedCanonical` and normalizing a bond.

src/MPS.jl
jofrevalles: It is a practical way to do that, just that

src/MPS.jl
mofeing: Yeah, my point is to have canonization and normalization separate, and have a `normalize!` method dispatched on `Canonical` that normalizes each $\Lambda$ independently.

So... exactly taking this line and moving it to `normalize!`.

src/MPS.jl
```diff
@@ -540,20 +541,29 @@ function mixed_canonize!(tn::AbstractMPO, orthog_center)
     return tn
 end
 
+LinearAlgebra.normalize(ψ::AbstractMPO; kwargs...) = normalize!(copy(ψ); kwargs...)
+LinearAlgebra.normalize(ψ::AbstractMPO, site::Site) = normalize!(copy(ψ), site)
```
mofeing, line 545: mmm maybe it's better if we write this method for `AbstractAnsatz`? that way we don't need to define it for all its subtypes.
Also, I think the first method is unnecessary because I believe (need to check) is already implemented for `AbstracTensorNetwork` or `AbstractQuantum`.

jofrevalles: No, `LinearAlgebra.normalize` it is not defined anywhere else.

mofeing: uuu then we should define it more generically for `AbstractQuantum` or so

src/MPS.jl
```diff
@@ -540,20 +541,29 @@ function mixed_canonize!(tn::AbstractMPO, orthog_center)
...
+LinearAlgebra.normalize!(ψ::AbstractMPO, site::Site) = normalize!(form(ψ), ψ; at=site)
 
 function LinearAlgebra.normalize!(::NonCanonical, ψ::AbstractMPO; at=Site(nsites(ψ) ÷ 2))
     tensor = tensors(ψ; at)
     tensor ./= norm(ψ)
     return ψ
 end
 
-LinearAlgebra.normalize!(ψ::AbstractMPO, site::Site) = normalize!(mixed_canonize!(ψ, site); at=site)
+LinearAlgebra.normalize!(config::NonCanonical, ψ::AbstractMPO; at=site) = normalize!(MixedCanonical(at), ψ; at)
```
mofeing, line 556: isn't this cheating? like you need to mixed canonize in order to use the `MixedCanonical` `normalize!` method

mofeing: you can always use the non-`Site` normalization for `NonCanonical`

jofrevalles: No, because inside the `normalize!(::MixedCanonical, ...)` there is already an step that does `mixed_canonize!`... I agree this is confusing. This step should be fixed in some way, since if the state is already in the `MixedCanonical` form, then it should not be necessary to do a whole `mixed_canonize!`. But for now is okay I guess.

mofeing: yes, but that `mixed_canonize!` inside the `normalize!(::MixedCanonical)` method dispatches to `mixed_canonize!(::MixedCanonical)` so it just moves the orthogonality center and doesn't call `canonize_site!` on the rest of the sites. so, if you call `mixed_canonize!` here, orthogonalizing to `at` site, then the interior `mixed_canonize!` will do nothing and that's completely fine.

so the original implementation is not more costly than this.

jofrevalles: Nevertheless, I noticed we had these two lines:
```julia
function LinearAlgebra.normalize!(::NonCanonical, ψ::AbstractMPO; at=Site(nsites(ψ) ÷ 2))
    tensor = tensors(ψ; at)
    tensor ./= norm(ψ)
    return ψ
end
LinearAlgebra.normalize!(ψ::AbstractMPO, site::Site) = normalize!(mixed_canonize!(ψ, site); at=site)
```
So maybe we can just use the first one, and remove the second for only `MixedCanonical` forms

jofrevalles:
@mofeing I refactored the `normalize` functions, so now we can do `normalize!` of an `MPS` in the `Canonical` form without recanonizing. We can also choose the `bond` at which it normalizes it.

src/MPS.jl
mofeing: > But yeah we should do that at some point. I don't think that is a particular a priority now, right? Maybe a comment will do

yeah, leave the original implementation commented and open an issue for implemented `mixed_canonize!(::MixedCanonical)` plus the reference to this code so it's changed when this is implemented

> So maybe we can just use the first one, and remove the second for only MixedCanonical forms

mmm fine for me

mofeing:
> @mofeing I refactored the `normalize` functions, so now we can do `normalize!` of an `MPS` in the `Canonical` form without recanonizing. We can also choose the `bond` at which it normalizes it.

nice! so I understand that if you don't provide the bond, then it normalizes all of them right? mmm do you have a use case for normalizing just 1 bond?

jofrevalles:
> nice! so I understand that if you don't provide the bond, then it normalizes all of them right? mmm do you have a use case for normalizing just 1 bond?

So I did not implemented it like that but I will change it so it is like you said. Maybe a use case would be when you truncate, right? Since you only truncate a bond and thus you want to renormalize the only "perturbed" bond.

jofrevalles:
I will do some more tests on `evolve!` this afternoon, for now let's keep this PR open. I think I can finish it today.

mofeing:
> So I did not implemented it like that but I will change it so it is like you said. Maybe a use case would be when you truncate, right? Since you only truncate a bond and thus you want to renormalize the only "perturbed" bond.

that could be a case. in general you would want such a method in the case that you want to normalize 1 $\Lambda$ and keep the rest of $\Lambda$s intact but I can't think of a real case where you would want that. if you want to normalize all the $\Lambda$s (or only one) calling `normalize!` on each $\Lambda$ shouldn't be costly because they're just vectors.

jofrevalles:
Okay for me this is ready to merge @starsfordummies @mofeing . I found that, in order to normalize all $\Lambda$, you need to divide all of them by `norm(mps)^(1/(L-1))`.

src/MPS.jl
```diff
@@ -540,20 +541,35 @@ function mixed_canonize!(tn::AbstractMPO, orthog_center)
...
+    if isnothing(at) # Normalize all λ tensors
+        normalizer = (norm(ψ))^(1 / (nsites(ψ) - 1))
+
+        for i in 1:(nsites(ψ) - 1)
+            λ = tensors(ψ; between=(Site(i), Site(i + 1)))
+            replace!(ψ, λ => λ ./ normalizer)
+        end
+    else
+        λ = tensors(ψ; between=at)
+        replace!(ψ, λ => λ ./ norm(ψ))
```
mofeing, line 571: shouldn't you just normalize the corresponding $\Lambda$ at that bond?
```suggestion
        λ = tensors(ψ; between=at)
        λ ./= norm(λ)
```

src/MPS.jl
```diff
@@ -540,20 +541,35 @@ function mixed_canonize!(tn::AbstractMPO, orthog_center)
...
-LinearAlgebra.normalize!(ψ::AbstractMPO, site::Site) = normalize!(mixed_canonize!(ψ, site); at=site)
-
 function LinearAlgebra.normalize!(config::MixedCanonical, ψ::AbstractMPO; at=config.orthog_center)
     mixed_canonize!(ψ, at)
     normalize!(tensors(ψ; at), 2)
     return ψ
 end
 
-# TODO function LinearAlgebra.normalize!(::Canonical, ψ::AbstractMPO) end
+function LinearAlgebra.normalize!(config::Canonical, ψ::AbstractMPO; at=nothing)
```
mofeing, line 561: `at` should be used for single sites, for bonds use the kwarg name `bond` so we keep consistency with other methods
```suggestion
function LinearAlgebra.normalize!(config::Canonical, ψ::AbstractMPO; bond=nothing)
```

src/MPS.jl
```diff
@@ -540,20 +541,35 @@ function mixed_canonize!(tn::AbstractMPO, orthog_center)
...
+    if isnothing(at) # Normalize all λ tensors
+        normalizer = (norm(ψ))^(1 / (nsites(ψ) - 1))
```
mofeing, line 563: mmm this seems a lil bit weird... is there a `norm(::Canonical)` method? can't you just call `normalize!(λ)` for each λ?

that exponentiation by 1/n is gonna give a lot of numerical problems for mid-large amount of sites

src/MPS.jl
```diff
@@ -540,20 +541,35 @@ function mixed_canonize!(tn::AbstractMPO, orthog_center)
...
+
+        for i in 1:(nsites(ψ) - 1)
+            λ = tensors(ψ; between=(Site(i), Site(i + 1)))
+            replace!(ψ, λ => λ ./ normalizer)
```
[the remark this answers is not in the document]
jofrevalles: Ah! Didn't know that, okay!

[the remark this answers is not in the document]
jofrevalles: Okay, after a long try, I found a bug related to that haha. I will post it as an issue but for now I will keep my implementation.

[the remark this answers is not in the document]
jofrevalles: I mean, I have been thinking about it and I think that is not a bug. This can not be done because if you have a `copy` of a `tn` (or if you do `normalize`), then you would modify all the inner arrays. I have tested it and that happens. What do you think @mofeing ?

jofrevalles:
Okay, this was complicated... @mofeing @starsfordummies I finally managed to make everything work. For me, this makes sense. At least now, I think tomorrow I will forget for sure... Let's try to write it here: 
`evolve!` has a `renormalize` kwarg, that for `NonCanonical` or `MixedCanonical` passes through `truncate`. This can not be done for `Canonical`, since `simple_update_2site!(::Canonical, ...)` contracts the two-site wave-function, evolves by calling  `simple_update_2site!(::NonCanonical, ...)`, and the `truncate` is there, but we need to truncate after that, when in we recover the (pseudo)`Canonical` form, and then is when we have to `normalize!` the bond.

src/Ansatz.jl
```diff
@@ -419,9 +426,16 @@ function simple_update_1site!(ψ::AbstractAnsatz, gate)
     return contract!(ψ, contracting_index)
 end
 
+function simple_update_2site!(
+    ::MixedCanonical, ψ::AbstractAnsatz, gate; threshold=nothing, maxdim=nothing, renormalize=false
+)
+    return simple_update_2site!(NonCanonical(), ψ, gate; threshold, maxdim, renormalize)
+end
+
 # TODO remove `renormalize` argument?
```
mofeing, line 435: remove this comment

src/Ansatz.jl
```diff
@@ -321,8 +328,8 @@ end
 Truncate the dimension of the virtual `bond` of a [`Canonical`](@ref) Tensor Network by keeping the `maxdim` largest
 **Schmidt coefficients** or those larger than `threshold`, and then recanonizes the Tensor Network if `recanonize` is `true`.
 """
-function truncate!(::Canonical, tn::AbstractAnsatz, bond; threshold, maxdim, recanonize=true)
-    truncate!(NonCanonical(), tn, bond; threshold, maxdim, compute_local_svd=false)
+function truncate!(::Canonical, tn::AbstractAnsatz, bond; threshold, maxdim, recanonize=false, renormalize=false)
+    truncate!(NonCanonical(), tn, bond; threshold, maxdim, compute_local_svd=false, renormalize=renormalize)
```
mofeing, line 332:
```suggestion
    truncate!(NonCanonical(), tn, bond; threshold, maxdim, compute_local_svd=false, renormalize)
```

src/Ansatz.jl
```diff
@@ -305,14 +308,18 @@ function truncate!(::NonCanonical, tn::AbstractAnsatz, bond; threshold, maxdim,
...
     return tn
 end
 
-function truncate!(::MixedCanonical, tn::AbstractAnsatz, bond; threshold, maxdim)
+function truncate!(::MixedCanonical, tn::AbstractAnsatz, bond; threshold, maxdim, renormalize=false)
     # move orthogonality center to bond
     mixed_canonize!(tn, bond)
-    return truncate!(NonCanonical(), tn, bond; threshold, maxdim, compute_local_svd=true)
+
+    return truncate!(NonCanonical(), tn, bond; threshold, maxdim, compute_local_svd=true, renormalize=renormalize)
```
mofeing, line 322:
```suggestion
    return truncate!(NonCanonical(), tn, bond; threshold, maxdim, compute_local_svd=true, renormalize)
```

src/Ansatz.jl
```diff
@@ -305,14 +308,18 @@ function truncate!(::NonCanonical, tn::AbstractAnsatz, bond; threshold, maxdim,
     end
 
     slice!(tn, virtualind, extent)
+    sliced_bond = tensors(tn; bond)
+
+    renormalize && replace!(tn, sliced_bond => sliced_bond ./ norm(tn))
```
mofeing, line 313: add comment acknowledging that `normalize!` would be better but can give problems on copied TNs

src/Ansatz.jl
```diff
@@ -274,8 +274,11 @@ Truncate the dimension of the virtual `bond` of a [`NonCanonical`](@ref) Tensor
     - `threshold`: The threshold to truncate the bond dimension.
     - `maxdim`: The maximum bond dimension to keep.
     - `compute_local_svd`: Whether to compute the local SVD of the bond. If `true`, it will contract the bond and perform a SVD to get the local singular values. Defaults to `true`.
+    - `renormalize`: Whether to renormalize the state at the bond after truncation. Defaults to `false`.
 """
-function truncate!(::NonCanonical, tn::AbstractAnsatz, bond; threshold, maxdim, compute_local_svd=true)
+function truncate!(
+    ::NonCanonical, tn::AbstractAnsatz, bond; threshold, maxdim, compute_local_svd=true, renormalize=false
```
mofeing, line 280: rename `renormalize` to `normalize` to keep consistency with Julia

```suggestion
    ::NonCanonical, tn::AbstractAnsatz, bond; threshold, maxdim, compute_local_svd=true, normalize=false
```

src/MPS.jl
```diff
@@ -540,20 +541,38 @@ function mixed_canonize!(tn::AbstractMPO, orthog_center)
     return tn
 end
 
+LinearAlgebra.normalize(ψ::AbstractMPO, site) = normalize!(copy(ψ), site)
```
mofeing, line 544: move this method to `AbstractAnsatz`

src/MPS.jl
```diff
@@ -540,20 +541,38 @@ function mixed_canonize!(tn::AbstractMPO, orthog_center)
...
+
 LinearAlgebra.normalize!(ψ::AbstractMPO; kwargs...) = normalize!(form(ψ), ψ; kwargs...)
+LinearAlgebra.normalize!(ψ::AbstractMPO, site) = normalize!(form(ψ), ψ; at=site)
```
mofeing, line 547: use keyword dispatching with `@kwmethod` or put do a specialized version for `Base.AbstractVecOrTuple{T}` (check `length(bond) == 2`). It could also work for the `BondMixedCanonical` form.

mofeing (approved):

PR #256, opened 2024-11-20
Move documentation to Vitepress

mofeing:
You can see the preview here: https://bsc-quantic.github.io/Tenet.jl/previews/PR256/


mofeing:
@jofrevalles do you think we should merge this at it is and add docstrings in subsequent PRs? or should turn this into a big PR that fixes all docs?

mofeing:
> Mmm I think it would be cool to fix all docs here, if we can already do this without needing anything else

I can try to fix what's already in here like the plots, but for that I want #211 or similar to be merged.

For the docstrings, we will need more work.

Issue #245, opened 2024-11-11
Test against Reactant


mofeing:
Reactant is slowly getting more and more stable, and it's working now in both macOS and Linux. We should write some tests.

PR #258, opened 2024-11-21
fix `Array`-`TracedRArray` contraction and force dense representation of Yao gates

mofeing:
CC @Todorbsc


mofeing (commented):
in the end there are multiple fixes for the VQE to work with Reactant

jofrevalles (commented):
Can we have a small `testset` that just tests whether the results from `Reactant` and `Tenet` is the same?

mofeing:
> Can we have a small `testset` that just tests whether the results from `Reactant` and `Tenet` is the same?

not right now because we are on a rush and we also need to create the infrastructure for Reactant tests (there are no tests with Reactant here)

#245 tracks that issue

jofrevalles (approved):

PR #269, opened 2024-12-10
Use `Compat.@compat` to mark public functions not exported

### LINKED ISSUE #228: Mark public API with `public` keyword
Now that API stabilization is coming closer and closer and users are starting to run with some problems on what should be used or not, we should start marking public API with the new `public` keyword which landed in Julia 1.11.

https://docs.julialang.org/en/v1/base/base/#public

- Exported symbols are automatically public
- There are some symbols which we want them to be public (i.e. users should not be afraid of using them) but not export them automatically (i.e. you need to access by explicit import or `Tenet.whatever` access). Those are the ones we must mark them as public.

Compat.jl provides the `@public` macro which acts as a no-op in Julia 1.10

mofeing:
Closes #228


mofeing:
@jofrevalles @Todorbsc tests are failing because i stopped exporting a couple of deprecated functions. I can fix it now but let's discuss it in the next meeting so we choose which to export and which not.

jofrevalles (approved):

mofeing:
This PR has no effect on execution of the code or API, so I'm merging it.

PR #271, opened 2024-12-26
Refactor `inds(::Tensor)` to return `Tuple`

mofeing:
Current implementation of `inds(::Tensor)` is to return the `Vector{Symbol}`, but this has the problem that mutating will modify the one from the original tensor too.


jofrevalles (approved):

mofeing:
Currently, only Reactant is failing. Core tests are passing which should be the only ones affected by this PR, so I'm merging.

PR #272, opened 2024-12-26
Replace KeywordDispatch for an in-house implementation

mofeing:
KeywordDispatch.jl is no longer actively developed and doesn't propagate docs. With our own implementation of keyword-dispatching, we have more control over it and can fix the propagation of docs.

The idea is to do sth as `Core.kwcall`; i.e. Julia lowers calls to methods with keyword-arguments to calls to `Core.kwcall` where the 1st argument is a `NamedTuple` with the arguments, the 2nd argument is the type of the function and the rest are the positional arguments. In our case, we call that `kwcall` manually on the default method without a `kwcall` function (but I can reconsider), so we implement keyword-dispatched methods as a method of the same function but where the kwargs are a `NamedTuple` that is passed as the first argument.

For example, for `inds`, the replacement for `@kwdispatch` is
```julia
inds(tn::AbstractTensorNetwork; kwargs...) = inds(sort_nt(values(kwargs)), tn)
```

The default `@kwmethod` is replaced for
```julia
inds(::@NamedTuple{}, tn::AbstractTensorNetwork) = inds((; set=:all), tn)
```

And the `set`-kwarg method is now
```julia
function inds(kwargs::NamedTuple{(:set,)}, tn::AbstractTensorNetwork)
    tn = TensorNetwork(tn)
    if kwargs.set === :all
...
```

It looks a lil more verbose than before, but we can add some helpers. The nice thing is that we can now put docstrings to these keyword-methods and also point to them.

Note: `sort_nt` is a function that sorts the keys of a `NamedTuple`. It's used for making keyword-dispatch invariant to kwarg order.


jofrevalles (approved):

mofeing:
Currently, tests fail due to Reactant.

I've run core tests in local and they pass all, so I'm merging.

PR #282, opened 2025-01-12
Add new `Circuit`, `Gate`, `Lane`, `Moment` types

mofeing:
while thinking about adding new features around connectivty between `Quantum`s, `Ansatz` and so, i needed to add this functionality to better represent quantum circuits or TEBD decompositions.

@jofrevalles this should fix your problem with getting the gates from a circuit in order. the only thing left to do for you to work is to implement a graph traversing o graph walk algorithm on top of `Circuit`.

### notes
- `Gate` is nice. it's a `Tensor` with `Site` information, just like `Quantum` is a `TensorNetwork` with `Site` information.
- `Lane` and `Bond` are going to be used for helping `Ansatz`. i need them for the PEPS.
- `Moment` can also be used for stacking `Quantum`s in a new type i'm designing for representing compositions of TNs.

### to do
we probably need to change the name of `Moment` because it means other thing in quantum circuit libraries


mofeing:
in the end, this PR got bigger than what i wanted but everything in here was needed

there is really no breaking changes, except for...

- `Lattice` (and thus `Ansatz`) now uses `Lane`s and not `Site`
  - this fixes issues with canonization on `Operator`s like `MPO` or dual states
  - also it's better defined
- `tensors(; at)`, `tensors(; bond)`, `tensors(; between)` now requires you passing a `Lane`
  - `inds(; at)` still accepts `Site`
- `simple_update!` now only accepts `Gate`, and not a "not very-well defined" `Quantum`

i want these breaking changes to be before 0.8.0 so expect another PR where i better define `Ansatz` with mapping between `Lane` to `Tensor` and `Bond` to index.

@jofrevalles i'm merging directly because i'm lil bit on a hurry and need this to implement `PEPS`. this stuff is still not very documented because i don't want it to be public yet: it will be public after some features that can be added on v0.8.x, but meanwhile you can start using `Gate` and `Circuit` for your work

jofrevalles (approved):

PR #288, opened 2025-01-14
Parallelize `CI` by splitting Unit Tests and Integration Tests on different Github actions

### LINKED ISSUE #273: Parallelize unit and integration tests
The idea is to run unit and integration tests in different GitHub Action jobs. For that, both "test/runtests.jl" and the ".github/workflows/CI.yml" files have to be modified.

It should be quite easy and should cut by half test times on CI.

jofrevalles:
This PR resolves #273 by modifying `runtests.jl` and `CI.yml` so the unit tests and integration tests run on different Github Actions.


test/runtests.jl
```diff
@@ -6,36 +6,38 @@ using Adapt
...
-    include("integration/ChainRules_test.jl")
-    # include("integration/BlockArray_test.jl")
-    include("integration/Dagger_test.jl")
-    include("integration/Makie_test.jl")
-    include("integration/KrylovKit_test.jl")
-    include("integration/Quac_test.jl")
-    include("integration/ITensors_test.jl")
-    include("integration/ITensorNetworks_test.jl")
+if "integration_tests" ∉ ARGS
+    @testset "Unit tests" verbose = true begin
```
mofeing, line 10: could do it such that the default is to run everything but that if a certain env var is set, then run the corresponding subset?

you have an example here: https://github.com/EnzymeAD/Reactant.jl/blob/c664414990141ab04455c49a793a3b9144f2182b/test/runtests.jl#L42-L45

mofeing (commented):
this is awesome! thanks!

i suspect that precompiled packages are not getting correctly cached because every CI run is precompiling all test dependencies and takes around 5-7min. it might be a good moment to fix this.

test/runtests.jl
jofrevalles: Ah sure! I will do that now!

jofrevalles:
> i suspect that precompiled packages are not getting correctly cached because every CI run is precompiling all test dependencies and takes around 5-7min. it might be a good moment to fix this.

I guess we can cache for each Julia version, right? I suppose this is not possible for different Julia versions. But how can this be done, in a sequential way? Since right now all the Actions are deployed at the same time.

test/runtests.jl
```diff
@@ -1,13 +1,14 @@
...
 using Tenet
 using OMEinsum
-using Reactant
 using Adapt
 
 include("Utils.jl")
 
 const TENET_TEST_GROUP = lowercase(get(ENV, "TENET_TEST_GROUP", "all"))
 
+println("Running tests for group: $TENET_TEST_GROUP")
```
mofeing, line 10: if you use `@info` we can get nice colors in the logs

```suggestion
@info "Running tests for group: $TENET_TEST_GROUP"
```

jofrevalles: Sure! I think I will remove that, I am putting this now since I think this `ENV` is not properly set by `CI.yml`.

mofeing:
> I guess we can cache for each Julia version, right? I suppose this is not possible for different Julia versions. But how can this be done, in a sequential way? Since right now all the Actions are deployed at the same time.

so the cache works and should be caching for each CI job type (e.g. one cache for "unit tests" - "ubuntu" - x86_64 - Julia 1.11).

but i still don't why it takes so much time to precompile for testing.

mofeing:
also, you may want to use the `matrix` instead of writing different jobs:

https://github.com/EnzymeAD/Reactant.jl/blob/c664414990141ab04455c49a793a3b9144f2182b/.github/workflows/CI.yml#L30-L33

and then you pass the env var like this:

https://github.com/EnzymeAD/Reactant.jl/blob/c664414990141ab04455c49a793a3b9144f2182b/.github/workflows/CI.yml#L130

jofrevalles:
> also, you may want to use the `matrix` instead of writing different jobs:
> 
> https://github.com/EnzymeAD/Reactant.jl/blob/c664414990141ab04455c49a793a3b9144f2182b/.github/workflows/CI.yml#L30-L33
> 
> and then you pass the env var like this:
> 
> https://github.com/EnzymeAD/Reactant.jl/blob/c664414990141ab04455c49a793a3b9144f2182b/.github/workflows/CI.yml#L130

Yes, I changed that. I am already doing this.

.github/workflows/CI.yml
```diff
@@ -24,14 +24,17 @@ concurrency:
   cancel-in-progress: ${{ startsWith(github.ref, 'refs/pull/') }}
 jobs:
   test:
-    name: Julia ${{ matrix.version }} - ${{ matrix.os }} - ${{ matrix.arch }} - ${{ github.event_name }}
+    name: Julia ${{ matrix.version }} - ${{ matrix.test_group }} - ${{ matrix.os }} - ${{ matrix.arch }} - ${{ github.event_name }}
```
mofeing, line 27: lets change the order of this info so its sorted in a much relevant way. also the os, arch and github event name don't seem to give much information because they are always the same.

```suggestion
    name: ${{ matrix.test_group }} / Julia ${{ matrix.version }}
```

mofeing (approved):

PR #290, opened 2025-01-16
Fix tensor-number multiplication

Todorbsc:
This PR resolves a bug where the type `T` does NOT exist but it should be type `Tensor` in multiplication operation.

### DIFF
--- a/test/Numerics_test.jl
+++ b/test/Numerics_test.jl
@@ -16,16 +16,6 @@
         @test parent(D) ≈ parent(A) - permutedims(parent(B), (3, 1, 4, 2))
     end
 
-    @testset "tensor-number multiplication" begin
-        data = rand(2, 3, 4)
-        realnum = 2
-        complexnum = 2 + 2im
-
-        A = Tensor(data, (:i, :j, :k))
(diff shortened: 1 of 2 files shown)


jofrevalles (approved):

test/Numerics_test.jl
```diff
@@ -16,6 +16,16 @@
...
+    @testset "tensor-number multiplication" begin
+        data = rand(2, 3, 4)
+        realnum = 2
+        complexnum = 2 + 2im
+
+        A = Tensor(data, (:i, :j, :k))
+        @test data * realnum ≈ parent(A * realnum) ≈ parent(realnum * A)
+        @test data * complexnum ≈ parent(A * complexnum) ≈ parent(complexnum * A)
+    end
+
```
mofeing, line 28: this is just an alias to `contract`, can you put it inside the `contract` testset? name the testset "`*` alias"

mofeing (changes requested):

test/Numerics_test.jl
Todorbsc: Done! check it out

mofeing (approved):

PR #291, opened 2025-01-16
Fix Qiskit circuit conversion to Tenet

Todorbsc:
This PR resolves two bugs:
- Type `Circuit` instead of `Quantum` when converting a Qiskit circuit to Tenet (resolved in https://github.com/bsc-quantic/Tenet.jl/pull/300)
- When converting a Qiskit circuit to a tensor network, start from site number `1` instead of `0`


mofeing (approved):
@Todorbsc like in #292, since CI is broken due to Reactant... do you mind running this tests locally and telling me if all tests pass?

jofrevalles (approved):
Thanks @Todorbsc !

jofrevalles (approved):

ext/TenetPythonCallExt/Qiskit.jl
```diff
@@ -11,7 +11,7 @@ function Base.convert(::Type{Circuit}, ::Val{:qiskit}, pyobj::Py)
     circuit = Circuit()
 
     for instr in pyobj
-        gatelanes = map(x -> Lane(pyconvert(Int, x._index)), instr.qubits)
+        gatelanes = map(x -> Lane(pyconvert(Int, x._index) + 1), instr.qubits)
```
mofeing, line 14: I just have one question: are we sure that Qiskit only uses 1-dimensional coordinates for qubits?

Todorbsc: If your question is whether the `qubits` field is always a 1D array, then yes, I think so.

Todorbsc: This says a `list`, so yes.
https://docs.quantum.ibm.com/api/qiskit/qiskit.circuit.QuantumCircuit#qubits

PR #296, opened 2025-01-18
Fix infinite recursion on `ChainRules.rrule(contract)`

mofeing:
this should fix CI definitely


mofeing:
i took advantage and also fixed some yao tests

the CI fails currently because it fails to find the `CondaPkg.toml` when running `]test` so no Python packages are installed

PR #299, opened 2025-01-20
Add basic benchmarking workflow, create benchmark/ folder with bench script

starsfordummies:
Ideally this should create some benchmark data in `dev/bench/data.js`  in the `gh-pages` branch, as well as an accompanying index.html  - I have an updated index.html as well but I don't know how to make it such that the pr puts it in the gh-pages folder...

### DIFF
--- a/.github/workflows/Benchmark.yml
+++ b/.github/workflows/Benchmark.yml
@@ -1,0 +1,66 @@
+name: Tenet benchmarks with BenchmarkTools.jl
+on:
+  push:
+    branches:
+      - main
+
+permissions:
+  contents: write
+  deployments: write
+
+jobs:
+  benchmark:
+    name: Run tenet benchmark
+    runs-on: ${{ matrix.os }}
+    strategy:
+      fail-fast: false
+      matrix:
+        version:
+          - '1'
+        os:
+          - ubuntu-latest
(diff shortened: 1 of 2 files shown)


mofeing:
yeah, don't worry. we can later add it to the `gh-pages` (although i think we need an automated way).

let's start by adding some benchmarks. @starsfordummies @jofrevalles @Todorbsc what do you think about benchmarking the following?

- contraction of 2 `Tensor`s (matmul) for Float64 & ComplexF64. at least 3 different sizes (small, mid & large).
- contraction of 2 `Tensor`s (more complex pattern) with the same variations.
- MPS overlap for different number of sites (L=10,=30,=60) and maximum bond dimensions
  - for L=10, use 2,32
  - for L=30, use 2,128,1024
  - for L=60, use 128,1024
- MPS canonization from `NonCanonical` to `Canonical`, from `NonCanonical` to `MixedCanonical` and from `MixedCanonical` to `MixedCanonical` in another site
  - yet to decide the benchmark space

we can benchmark more stuff in the future, but let's start with the following which are key to us right now

mofeing:
@jofrevalles that sounds great! i would say that feel free to push to this PR so we can see the result

PR #300, opened 2025-01-20
Fix Python environment and integration tests for Python extensions

Todorbsc:
- Python environment is not being created with the required dependencies for the Tenet-Python extensions and Python integration tests cannot run properly
- Some Python integration tests gave errors when the environment issue was resolved


test/integration/python/test_cirq.jl
```diff
@@ -17,12 +17,12 @@ CondaPkg.update()
         circuit.append(cirq.H(qubits[1]))
         circuit.append(cirq.H(qubits[2]))
 
-        circ = convert(Circuit, circuit)
+        circ = convert(Circuit, Val(:cirq), circuit)
```
mofeing, line 20: no, this should be without the `Val` and it should be automatically dispatched.

test/integration/python/test_cirq.jl
```diff
@@ -37,13 +37,13 @@ CondaPkg.update()
...
```
mofeing, line 45: same as above, this should be called without the `Val`

test/integration/python/test_qibo.jl
```diff
@@ -11,12 +11,12 @@
     circuit.add(qibo.gates.H(1))
     circuit.add(qibo.gates.H(2))
 
-    circ = convert(Circuit, circuit)
+    circ = convert(Circuit, Val(:qibo), circuit)
```
mofeing, line 14: here too

test/integration/python/test_qiskit.jl
```diff
@@ -11,12 +11,12 @@
...
     circuit.h(2)
 
-    tn = convert(Circuit, circuit)
+    tn = convert(Circuit, Val(:qiskit), circuit)
     @test tn isa Circuit
-    @test issetequal(sites(circ; set=:inputs), Site.([1, 2, 3]; dual=true))
-    @test issetequal(sites(circ; set=:outputs), Site.([1, 2, 3]))
-    @test Tenet.ntensors(circ) == 7
+    @test issetequal(sites(tn; set=:inputs), Site.([0, 1, 2]; dual=true))
+    @test issetequal(sites(tn; set=:outputs), Site.([0, 1, 2]))
```
mofeing, line 17: didn't we decided last week to put a `+ 1` so it's 1-indexed?

mofeing (changes requested):

mofeing (approved):
everything good! thanks @Todorbsc!

remember to reenable all the other tests, and do a SQUASH MERGE

jofrevalles (approved):

PR #313, opened 2025-02-03
Update calls to deprecated methods

mofeing:
Deprecation warnings remain. This PR just fixes calls to deprecated methods on internal code and tests.

@jofrevalles I've renamed `contract!` on `MPS` for contracting the $\Lambda$ to a tensor to `absorb!` because (1) it was conflicting with `contract!(; bond)` and the dispatch wasn't working, and (2) I think semantically this is more representative. wdyt?

### DIFF
--- a/test/unit/MPO_test.jl
+++ b/test/unit/MPO_test.jl
@@ -100,10 +100,10 @@ end
             @test isisometry(canonized, Lane(i); dir=:right)
         elseif i == 5 # in the limits of the chain, we get the norm of the state
             normalize!(tensors(canonized; bond=(Lane(i - 1), Lane(i))))
-            contract!(canonized; bond=(Lane(i - 1), Lane(i)), dir=:right)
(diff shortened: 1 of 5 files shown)


jofrevalles:
Seems good to me!

mofeing:
Great, then let's merge once all tests pass.

test/unit/MPO_test.jl
```diff
@@ -1,3 +1,5 @@
+using Test
```
jofrevalles, line 1:
```suggestion
```
This is not needed, right?

jofrevalles (changes requested):

test/unit/MPO_test.jl
mofeing: this is a recent change i've done: now we include everything needed in the file so that we can run tests on just one file when debugging.

Also, for integration tests now we use SafeTestsets.jl which bundles everything in a module to avoid symbol clashes. In exchange, we need to add the `using` or `import` statements on each file.

PR #329, opened 2025-02-12
Support multiple backends for `contract`

mofeing:


jofrevalles:
This is great, @mofeing, thanks!

src/Numerics/Contract.jl
```diff
@@ -0,0 +1,138 @@
...
+
+    - `dims`: indices to contract over. Defaults to the set intersection of the indices of `a` and `b`.
+    - `out`: indices of the output tensor. Defaults to the set difference of the indices of `a` and `b`.
+
+!!! todo
+
+    We are in the process of making [`contract`](@ref) multi-backend; i.e. let the user choose between different einsum libraries as the engine powering [`contract`](@ref).
+    Currently, we use [OMEinsum.jl](@ref), but it has proven to be slow when used dynamically like we do.
+"""
+contract(a::Tensor, b::Tensor; kwargs...) = contract(default_backend[], a, b; kwargs...)
```
lkdvos, line 45: This is a duplicate implementation of L19

src/Numerics/Contract.jl
```diff
@@ -0,0 +1,138 @@
...
+    contract(a::Tensor; dims=∩(inds(a), inds(b)), out=nothing)
+
+Perform a unary tensor contraction operation.
+
+# Keyword arguments
+
+    - `dims`: indices to contract over. Defaults to the repeated indices.
+    - `out`: indices of the output tensor. Defaults to the unique indices.
+"""
+contract(a::Tensor; kwargs...) = contract(default_backend[], a; kwargs...)
```
lkdvos, line 57: This too

src/Numerics/Contract.jl
```diff
@@ -0,0 +1,138 @@
...
+
+"""
+    contract!(c::Tensor, a::Tensor, b::Tensor)
+
+Perform a binary tensor contraction operation between `a` and `b` and store the result in `c`.
+"""
+contract!(c::Tensor, a::Tensor, b::Tensor) = contract!(default_backend[], c, a, b)
```
lkdvos, line 64: Here you might also want to consider a definition like L19 that captures all of them?

ext/TenetReactantExt.jl
```diff
@@ -213,6 +214,10 @@ Base.@nospecializeinfer @noinline function Tenet.contract(
     return Tensor(data, ic)
 end
 
+@reactant_overlay function Tenet.contract(@nospecialize(::AbstractBackend), @nospecialize(xs::Tensor...); kwargs...)
+    return reduce((x, y) -> contract(x, y; kwargs...), xs)
+end
```
lkdvos, line 219:
```suggestion
@reactant_overlay function Tenet.contract(@nospecialize(b::AbstractBackend), @nospecialize(xs::Tensor...); kwargs...)
    return reduce((x, y) -> contract(b, x, y; kwargs...), xs)
end
```
I think you might have lost the backend here

lkdvos (commented):
Hi Sergio, this definitely looks nice!

It's a bit strange to me to not immediately use the `AbstractBackend` in TensorOperations, and instead defining your own, which then wraps objects of the exact same name, but that's up to you 😉 

Considering the type stability etc,  I guess we discussed this already, but my guess would be that the incurred stability loss from dynamically determining the contraction (overlapping the indices etc) already makes you do dynamic dispatch at every call to `contract`, since the `ndims` of all of the arrays are not compile-time inferrable. The upshot is ofcourse that otherwise every different contraction pattern possibly has to be recompiled.

I do think that determining the backend once and passing it to all pairwise operations is a good idea, but keep in mind that you would need a function barrier to restore the type stability there as well :)

In any case, overall looks like a nice improvement :)

ext/TenetReactantExt.jl
mofeing: Yes, that's on purpose because when tracing with Reactant.jl, it uses no backend. On tracing, Reactant.jl just registers that a contraction is performed (i.e. by emitting the `stablehlo.dot_general` MLIR op) but no information of the backend is set. Afterwards, XLA will choose the correct implementation for the backend we want (i.e. CPU, GPU, TPU or distributed).

tldr: Reactant.jl ignores the "backend" and sets its own.

test/integration/TensorOperations_test.jl
```diff
@@ -0,0 +1,121 @@
...
+    # NOTE pattern not supported by TensorOperations
+    # @testset "diagonal" begin
+    #     A = Tensor(rand(2, 3, 2), (:i, :j, :i))
+
+    #     C = contract(A; dims=())
+    #     C_ein = ein"iji -> ij"(parent(A))
+    #     @test inds(C) == (:i, :j)
+    #     @test size(C) == size(C_ein) == (2, 3)
+    #     @test parent(C) ≈ C_ein
+    # end
```
jofrevalles, line 29: Why is this commented out?

mofeing: TensorOperations doesn't support these patterns. You can find the full discussion in #208

ext/TenetReactantExt.jl
mofeing: Actually, you do a pretty good job and Reactant.jl is slightly slower than TensorOperations.jl, so well done guys :)

but in any case, Reactant.jl is a compiler pipeline and its power is on seeing *all* the contractions collectively, and optimizing all the contractions as a whole.

one cool idea we are working on in Reactant.jl is "kernel injection" (right now only supported for CUDA.jl kernels but will add it support for generic Julia in the future): i.e. you can write a kernel in Julia, inject the LLVM IR in our MLIR and we compile it in such a way that XLA can still compile it.

it would be quite cool to check if we can optionally inject TensorOperations into it, but that will be a future project.

lkdvos:
Not sure if this helps, but at least for TensorOperations the type stability is only guaranteed for the signatures with the tuples of Int. (For the not-in-place versions). Because we know the length of the tuples, we know the ndims of the output at compile time.

PR #333, opened 2025-02-17
Distributed TEBD library

mofeing:
@jofrevalles we can start adding code for the distributed TEBD here


jofrevalles:
I prefer to do the PR once everything is working. For now, I am using my branches and I will keep modifying the code as we discussed

mofeing:
sure, i'm just adding the infrastructure so you can build from here as you want.

PR #338, opened 2025-03-06
Avoid computing the `norm` on normalization of a `Canonical` MPS

mofeing:
This should speed up `normalize!` on a `Canonical` MPS


mofeing:
All tests pass both in Julia 1.10 & 1.11 on my laptop... don't know why it fails on GitHub runners.

jofrevalles (approved):

PR #340, opened 2025-03-07
Move Python integration tests to their own folder

mofeing:
CI is giving a lot of caching problems due to CondaPkg: it doesn't cache the Python environment so it takes a long time to run any test because is reresolving the environment and installing everything... again.

This PR moves Python tests to another folder to clean unit and integration tests and alleviate caching issues.


.github/workflows/CI.yml
```diff
@@ -63,6 +63,52 @@ jobs:
...
         with:
           files: lcov.info
+  python:
+    name: ${{ matrix.test_group }} / Julia ${{ matrix.version }}
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        version:
+          - '1'
```
jofrevalles, line 73:
```suggestion
          - '1.11'
```
Isn't that what we want?

mofeing: just setting '1' is ok as it means "get me the latest 1.x version"

jofrevalles (approved):

Issue #356, opened 2025-06-26
access MPS tensors with custom getindex


starsfordummies:
for an MPS where the ordering is clear it's much easier if we can 
```
function Base.getindex(x::MPS, i)
           tensor_at(x,CartesianSite(i))  # ideally site"$(i)" soon
       end
```

should we also implement replace_tensor() as setindex ?

mofeing:
Fixed in Tangles.jl

PR #357, opened 2025-06-26
Getsetindex mps

starsfordummies:
related issue #356

for getindex and length it looks pretty straightforward

for setindex there's the usual freedom to break stuff: do we want to put restrictions or let it open for all unsafe play  ? ie. replacing a tensor with a tensor with a different number of indices, with different sizes, with different indices, etc... 

In its current form it allows in principle also to append a tensor to the end, but of course there also one should correct the second-last tensor to get the additional bond, so it's not so trivial . open to feedback


mofeing:
Nice! Don't forget to run the formatter.

PR #360, opened 2025-07-04
Added a basic ```autoMPO``` formulation

nbaldelli:
I added a basic formulation of an automatic MPO engine. It accepts single and two-site operators at arbitrary distances.
It accepts as arguments a list of tuples with the sites where the operators are applied, the operators and the corresponding prefactors.

For example, to write down an Hamiltonian like

$$
-\sum_{i} \sigma^x_i \sigma^x_{i+1} - \sum_{i} \sigma^z_i \sigma^z_{i+2} -0.5 \sum_{i} \sigma^x_i.
$$

one can pass 

```
one_body = [(1, X, -0.5)]
two_body = [(1, 2, sx, sx, -1.0),(1, 3, sz, sz, -1.0)]

W = autoMPO_periodic(L, one_body, two_body),
```

and the sites are implicitly assumed to be infinitely repeating. The MPO matrices are then projected to vectors at sites 1 and L for a finite chain. I can add a test to show that for example reproduces the correct Ising MPO. I also have an eager autoMPO formulation (does not assume translational invariance and constructs separate matrices at each site, at cost of a larger bond dimension), but I have to check it better.

### DIFF
--- a/src/Operations/simple_update.jl
+++ b/src/Operations/simple_update.jl
@@ -19,8 +19,22 @@ end
 
 function generic_simple_update!(tn, operator; maxdim=nothing)
     op_sites = acting_sites(operator)
-    @assert length(op_sites) == 2 "Operator must have exactly two sites"
+    @assert 1 <= length(op_sites) <= 2 "Operator must act on one or two sites"
     @argcheck all(Base.Fix1(hasplug, tn), Plug.(op_sites; isdual=false)) "Operator plugs must be present in the MPS"
+
+    # TODO move to function?
+    # TODO do not move orthogonality center if 1-site... and unitary?
+    # shortcut for 1-site operator
+    if length(op_sites) == 1
+        _site = only(op_sites)
+        _tensor = tensor_at(tn, _site)
+        tmp_contracting_ind = Index(gensym(:tmp))
+        tensor = replace(_tensor, ind_at(tn, plug"$_site") => tmp_contracting_ind)
+        operator = replace(operator, Index(plug"$_site'") => tmp_contracting_ind)
+        new_tensor = Muscle.binary_einsum(tensor, operator)
+        replace_tensor!(tn, _tensor, new_tensor)
+        return tn
+    end
 
     site_a, site_b = minmax(op_sites...)
     old_tensor_a = tensor_at(tn, site_a)
@@ -61,6 +75,8 @@ function generic_simple_update!(tn, operator; maxdim=nothing)
 
     return tn
 end
+
+simple_update!(tn::AbstractTangle, operator::Tensor; kwargs...) = generic_simple_update!(tn, operator; kwargs...)
 
 ## `MPS`
 function simple_update!(tn::MPS, operator::Tensor; kwargs...)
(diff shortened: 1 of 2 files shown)


jofrevalles:
Nice! Thanks @nbaldelli

src/Components/MPO.jl
```diff
@@ -160,3 +160,53 @@ function Base.rand(rng::Random.AbstractRNG, ::Type{MPO}; n, maxdim=nothing, elty
...
+    H = sum_i    c1_i * O_i
+    + sum_{i<j} c2 * O_i ⊗ O_j
+where O_i and O_j are local operators acting on sites i and j, respectively.
+
+#Arguments
+  - `L` : length of the MPO
+  - `one_body` : list of (i, O, c1)            single‐site terms
+  - `two_body` : list of (i, j, O_i, O_j, c2)  two‐site terms
+"""
+function autoMPO_periodic(L, one_body, two_body; type = ComplexF64)
```
mofeing, line 175: Mind changing the name to `autompo_periodic`? The convention is that functions are in snake_case, types and traits are in camelCase.

```suggestion
function autompo_periodic(L, one_body, two_body; type = ComplexF64)
```

src/Components/MPO.jl
```diff
@@ -160,3 +160,53 @@ function Base.rand(rng::Random.AbstractRNG, ::Type{MPO}; n, maxdim=nothing, elty
...
+        finish = next_chan + d - 1
+        next_chan = finish + 1
+
+        @views W[1,start,:,:] .+= Oi #Insertion of first operator
+
+        for k in i+1:j-1
+            @views W[start + (k-(i+1)),start + (k-(i+1)) + 1,:,:] .+= Id #Propagation of identities through unaffected sites
+        end
+        @views W[finish,D,:,:] .+= c2 .* Oj #Insertion of second operator
+    end
```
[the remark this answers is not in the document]
nbaldelli: No, it shouldn't be a problem because different c2 terms write on different blocks of ```W```. If you have an Hamiltonian for which the c2 related to some coupling changes site by site (so something like $$c2_{ij} X_i X_j$$), this formulation does not work anymore as the system cannot be considered translationally invariant.

src/Components/MPO.jl
```diff
@@ -160,3 +160,58 @@ function Base.rand(rng::Random.AbstractRNG, ::Type{MPO}; n, maxdim=nothing, elty
     # TODO order might not be the best for performance
     return MPO(arrays; order=(:l, :i, :o, :r))
 end
+
+"""
+    autompo_periodic(L, one_body, two_body; eltype=ComplexF64)
+
+Builds a translationally invariant MPO of length ``L`` for Hamiltonians of the form:
+
+H = sum_i alpha_i * O_i + sum_{i<j} beta_{ij} * O_i ⊗ O_j
```
mofeing, line 169: The double tick is required to be formatted with Latex, or are you avoiding that?

```suggestion
``H = \sum_i alpha_i * O_i + \sum_{i<j} beta_{ij} * O_i \otimes O_j``
```

mofeing (approved):
There are a couple of formatting issues, but I can fix them after merging the PR.

Image

mofeing (approved):

PR #384, opened 2025-08-06
expectation value of a one-site operator for MPS using mixedcanonical form

starsfordummies:
probably not too tenet-idiomatic but should work

### DIFF
--- a/src/Components/MPS.jl
+++ b/src/Components/MPS.jl
@@ -212,9 +212,7 @@ function Base.show(io::IO, psi::MPS)
     print(io, "MPS (#tensors=$(ntensors(psi)), #inds=$(ninds(psi)), maxbondsize=$(maxbondsize(psi)))")
 end
 
-
-""" Expectation value for a one-site operator `op` on site `op_site` of mps `psi`, brings to canonical form. 
-This makes a copy of the MPS """
+""" Expectation value for a one-site operator `op` on site `op_site` of mps `psi`, brings `psi` to canonical form. """
 function expect_1site!(psi::MPS, op::AbstractArray, op_site::Int)
         canonize!(psi, MixedCanonical(site"$(op_site)"))
         plug_ind = ind_at(psi, plug"$(op_site)")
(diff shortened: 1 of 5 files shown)


jofrevalles (commented):
This can work well, but maybe is better if we specialize this function for `MixedCanonical` form, and the default `MPS` expect function calls the one of `MixedCanonical` with `canonize(MPS)`?

starsfordummies:
yes, definitely! I just wanted to get the discussion started, since we definitely need a function to compute exp values after it disappeared in the latest refactorings

src/Components/MPS.jl
```diff
@@ -209,5 +209,23 @@ end
...
+
+        return ev
+end
+
+""" Expectation value for a one-site operator `op` on site `op_site` of mps `psi`.
+This makes a copy of the MPS """
+function expect_1site(psi::MPS, op::AbstractArray, op_site::Int)
+        psi = copy(psi)
+        expect_1site!(psi, op, op_site)
 end
```
mofeing: mind moving them to a `Operations/expect.jl` file please?

src/Operations/evolve.jl
```diff
@@ -11,6 +11,11 @@ struct ZipUpAlgorithm <: EvolveAlgorithm end
 
 function evolve! end
 
+function evolve(tn, op; kwargs...)
+    tn = copy(tn)
+    evolve!(tn, op; kwargs...)
+end
+
```
mofeing: can be done in one line


```suggestion
evolve(tn, op; kwargs...) = evolve!(copy(tn), op; kwargs...)
```

mofeing (commented):
thanks for starting the work on this @starsfordummies 

@jofrevalles yeah, the idea is to dispatch based on the CanonicalForm so that we can specialize behavior

the specialized implementations should do the following

- on `NonCanonical`, i would just do a sweep contracting the tensors
- on `MixedCanonical`,
  1. `canonize!` to the range of sites of the operator
  2. directly connect boundary virtual bonds between ket and bra (skipping the tensors outside of the range of the operator)
  3. call `contract`
- on `BondCanonical`, `canonize!` to the range of sites of the operator (so move to `MixedCanonical`) and call `expect` again but on the `MixedCanonical` form
- on `VidalGauge`, i would directly take the range of sites of the operator (so move to `MixedCanonical`), absorb the lambdas and contract

if you know a more efficient way, we can do it that way

starsfordummies:
i added a few methods for exp values of local ops and mpo - it could well be that bringing the MPS to canonical form is always the best way (even if it involves making a copy+canonizing it), need to benchmark it

